Issue #12, opened 2018-06-25
Possible incorrect output from `uriUriStringToUnixFilenameA()`


vs49688:
I have a Java program that makes some URIs and sends them to a C++ program (using uriparser) to something with. Below is an example of what Java's giving me:

```bash
ubuntu@DEV:~$ jshell
|  Welcome to JShell -- Version 10.0.1
|  For an introduction type: /help intro

jshell> Path p = Paths.get("/home/ubuntu/")
p ==> /home/ubuntu

jshell> p.toUri()
$2 ==> file:///home/ubuntu/

jshell> p.resolve("somefolder/").toUri()
$3 ==> file:///home/ubuntu/somefolder/

jshell> p.toUri().resolve("somefolder/")
$4 ==> file:/home/ubuntu/somefolder/
```

I'm running under the assumption that Java is doing the right thing here. Putting those URIs into `uriparse`:

```bash
ubuntu@DEV:~/Desktop/urib/build$ ./uriparse file:///home/ubuntu/somefolder/
uri:          file:///home/ubuntu/somefolder/
scheme:       file
hostText:     
 .. pathSeg:  home
 .. pathSeg:  ubuntu
 .. pathSeg:  somefolder
 .. pathSeg:  
absolutePath: false
              (always false for URIs with host)
```

```bash
ubuntu@DEV:~/Desktop/urib/build$ ./uriparse file:/home/ubuntu/somefolder/
uri:          file:/home/ubuntu/somefolder/
scheme:       file
 .. pathSeg:  home
 .. pathSeg:  ubuntu
 .. pathSeg:  somefolder
 .. pathSeg:  
absolutePath: true
```

Now, for the actual issue:

If I stick those URIs into `uriUriStringToUnixFilenameA()`, I get:
  * `file:///home/ubuntu/somefolder/` -> `/home/ubuntu/somefolder/`, and
  * `file:/home/ubuntu/somefolder/` -> `file:/home/ubuntu/somefolder/`

My other code chockes on the second output as it's expecting a valid path.

Does this look like a bug in uriparser, or am I simply using it wrong?

I've attached a sample program [here](https://gist.github.com/vs49688/583f1f6d7824493f75beedfae7dfdbc3) that shows this behaviour.

If this is indeed a bug I have a patch ready, but I'd want to be sure first.

EDIT: This is using current master (20ae776ab86bf776c97086064a1f6a78bffcabfd)

hartwork:
[...]

PR #13, opened 2018-06-28
Fixed handling of file: URIs with no authorities.

vs49688:
Fix for #12


hartwork:
Hi @vs49688 ,

I have god news and bad news. I hope you're not too disappointed but during review I noticed that I wanted a fix that was less invasive to the existing code base and closer to the approach that I already had in mind for a fix. So I made my own version. Would you mind to have a look at #14?

Thanks and best, Sebastian

PR #14, opened 2018-06-30
Proposed fix for issue #12

hartwork:
PS: @vs49688 I'm happy to put your real name in for credit instead, if that's something you're willing to share.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -2,6 +2,10 @@
 
   * Fixed: Documentation typos (GitHub #10, #11)
       Thanks to Graham Percival!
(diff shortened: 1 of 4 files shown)


vs49688:
So I just had a look, they're practically the same fix except that I separated that behemoth of a ternary statement. This is definitely neater, so I won't be offended if you don't use mine!

As for my real name, feel free. I'd rather not say it on a public issue tracker, but you should have it from our email exchange when you first moved uriparser to Github.

hartwork:
Glad to hear. Thanks!

Issue #17, opened 2018-08-01
Migrate to CMake / Overly complicated Windows building


computerquip-streamlabs:
I'd like to use this in my own project but getting this running on Windows hasn't proven obvious and there are no instructions (outside of the possibility of just adding all the source files and include files into my own project which might be intended) .

Is there any chance I could convince you to adopt a CMake/Meson/\<literally anything else other than autotools\> for a build system?

hartwork:
Pesonally I "like" GNU Autotools and never really "liked" CMake but I get that CMake does a lot more for us with targetting non-Unix systems so it's only a matter of time until Autotools alone (or at all) stops sufficing.

hartwork:
Fixed by merge of #47

For any potential issues with the CMake integration, please open new tickets. Thank you!

PR #18, opened 2018-08-01
Experimental - CMake build system

computerquip-streamlabs:
In response to https://github.com/uriparser/uriparser/issues/17, I've implemented a slightly unfinished port to the CMake build system. This supports a few nifty things already however:

* Install rules, including cpack. Using cpack, you can create various distribution archives and types, including zip files, rpm packages, etc. This helps with distributing across the different platforms in general. The installation structure still follows GNU style, including generating the pkg-config file.

* Works with virtually any toolset you can throw at CMake, potentially including all of [these](https://github.com/ruslo/polly), as long as the code can compile with that toolset.

* Significantly smaller and simple. It's easier to read and maintain in general as a result.

This isn't finished though! I still need to integrate CppTest and test on non-Windows platforms (and just test in general). This is mostly to give an idea what a not-autoconf build system would look like for the project.


hartwork:
PS: CMake-based AppVeyor integration will be nice to have, too

hartwork:
PS: I have plans on migrating from cpptest to googletest so in case the cpptest buts turn out to be a bigger problem than googletest bits would be, we can have it depend on that to happen first, if you like

markand:
superseded by #23, this one should be closed.

PR #41, opened 2018-12-05
Save 1 byte of waste in uriToStringCharsRequired with URIs containing IPv4 addresses

gyh007:
fix space requirements were 1 byte bigger than necessary when host is ipv4 address in uriToStringCharsRequired*

### DIFF
--- a/src/UriRecompose.c
+++ b/src/UriRecompose.c
@@ -246,7 +246,7 @@ static URI_INLINE int URI_FUNC(ToStringEngine)(URI_CHAR * dest,
 									}
 								}
 							} else {
-								(*charsRequired) += charsToWrite + 1;
+								(*charsRequired) += charsToWrite + ((i == 3) ? 0 : 1);
 							}
 						}
 					} else if (uri->hostData.ip6 != NULL) {


hartwork:
Hi @gyh007,

interesting patch, looks "right" to me right now. I'll need to sleep over whether this change "requires" a test case and if I want that extra complexity for saving one byte. I just checked: `uriToStringCharsRequired` is _not_ the one assuming worst case needs (to still be performant) — it's `uriComposeQueryCharsRequiredEx` — so it seems to be exact to the byte everywhere else (which argues for applying that patch).

Out of curiosity, do you have a scenario where that single byte counts? Happy to read offline if online does not fit.

Best, Sebastian

hartwork:
PS: The commit message will need better text to be merge 1:1. How about something like `Stpo wasting 1 byte in uriToStringCharsRequired with IPv4 URIs`?

hartwork:
I see. Something like ..
```
std::string buffer(size);
uriToStringA((char *) buffer.data(), &dest, size + 1, &size);
buffer[size] = '\0';
```
.. would not have that problem. Personally, I'd always write the terminator as a last step for zero chance that something else overwrote it after.

What do you think?

gyh007:
`
buffer[size] = '\0'; 
`
won't help, because i doesn't change the size of the buffer, the terminator won't end the std::string, it is a valid char. `buffer.resize(size-1)` should work.

hartwork:
I played a bit with this myself now. Calling `.resize` after seems like the right thing to me (independent of changes to `uriToStringCharsRequired`):

```c
#include <iostream>
#include <cstring>

using namespace std;

void dump(const string & buffer) {
    cout
        << "buffer=\"" << buffer << "\""
        << "  buffer.data=\"" << buffer.data() << "\""
        << "  buffer.size=" << buffer.size()
        << "  strlen(buffer.data)=" << strlen(buffer.data())
        << endl;
}

int main() {
    string buffer("", 20);
    dump(buffer);

    strcpy(const_cast<char *>(buffer.data()), "123456789");
    dump(buffer);

    buffer.resize(5);
    dump(buffer);

    return 0;
}
```

```console
# g++ -Wall -Wextra main.cxx && ./a.out
buffer="T       �����"  buffer.data=""  buffer.size=20  strlen(buffer.data)=0
buffer="123456789       �����"  buffer.data="123456789"  buffer.size=20  strlen(buffer.data)=9
buffer="12345"  buffer.data="12345"  buffer.size=5  strlen(buffer.data)=5
```

hartwork:
Merged no, thank you!

Your commit (with message changed): fe138783223bc0a4f247fb74b91f204dd7251522
New tests at 047bffdb7262677546bee827f23f3e4315f3742c and caa1f980c178de9553458af56ba360ffa103ba0a.

PR #45, opened 2019-01-06
Fix Haiku build, link to -lsocket for inet_ntop

Begasus:
Haiku fails the build when not linked against -lsocket/-lnetwork, this would reduce the need of creating a patch for Haiku and shouldn't affect other OS's

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -26,6 +26,8 @@ AM_CONDITIONAL([WIN32], test "$WIN32" = "yes")
 AC_CHECK_FUNCS([wprintf])
 AC_CHECK_FUNCS([reallocarray])  # needs AC_USE_SYSTEM_EXTENSIONS (#define _GNU_SOURCE)
 
+## Haiku check for -lsocket -lnetwork inet_ntop
+AC_SEARCH_LIBS([inet_ntop],[socket network])
 
 ## Option --disable-test
 AC_ARG_ENABLE(test, [


configure.ac
```diff
@@ -26,6 +26,8 @@ AM_CONDITIONAL([WIN32], test "$WIN32" = "yes")
 AC_CHECK_FUNCS([wprintf])
 AC_CHECK_FUNCS([reallocarray])  # needs AC_USE_SYSTEM_EXTENSIONS (#define _GNU_SOURCE)
 
+## Haiku check for -lsocket -lnetwork inet_ntop
+AC_SEARCH_LIBS([socket],[socket inet network])
```
hartwork: The [signature of `AC_SEARCH_LIBS`](https://www.gnu.org/software/autoconf/manual/autoconf-2.66/html_node/Libraries.html) is …
> `AC_SEARCH_LIBS (function, search-libs, [action-if-found], [action-if-not-found], [other-libraries])`

… so the use above would pass `socket` for parameter `function`.
Could it be we want `inet_ntop` for `function`, instead?

hartwork: Also, is `inet` needed? The comment line implies otherwise, in a way.

Begasus: Need to review this, I'm not a developer so I used a base from another recipe, if you have a better suited solution I'm happy to hear about it :) I'll get back on this when I have the time again to look more into it

hartwork: Not a better solution, no. I would need a running Haiku to debug this myself, just noticed that there is a difference between comment and implementation. Please keep me updated, thank you!

hartwork: It "looks better" but does it fix the problem? I cannot tell.

Begasus: Just checked, on gcc2hybrid (my main Haiku install):
For gcc2 it's using -lsocket
For gcc7 it's using -lnetwork to link against
Both are working ok to build and create the package and passes the tests (although I don't see any actual test being performed with "make check"?)

hartwork: > For gcc2 it's using -lsocket
> For gcc7 it's using -lnetwork to link against

Sounds excellent, thank you!

> (although I don't see any actual test being performed with "make check"?)

`make check` should be running tests. The code depends on C++98 and GTest >=1.8.1. Maybe some part of the Haiku packaging is cutting that code away for you?

hartwork: PS: I'll need to convert this to CMake for #47 soon it seems. Where can I find the current version of CMake in Haiku?

hartwork: PS: But let's first merge the Autotools version...

hartwork: PS: My current CMake working draft at #47 is this:
```cmake
if(HAIKU)
    # Function inet_ntop needs -lsocket or -lnetwork (see pull request #45)
    check_library_exists(socket inet_ntop "" HAVE_LIBSOCKET__INET_NTOP)
    check_library_exists(network inet_ntop "" HAVE_LIBNETWORK__INET_NTOP)
    if(HAVE_LIBSOCKET__INET_NTOP)
        target_link_libraries(uriparse PUBLIC socket)
    endif(HAVE_LIBSOCKET__INET_NTOP)
    if(HAVE_LIBNETWORK__INET_NTOP)
        target_link_libraries(uriparse PUBLIC network)
    endif(HAVE_LIBNETWORK__INET_NTOP)
endif(HAIKU)
```

Begasus: I'll  look into it today :)

Begasus:
Not sure why the check from travis failed here?

hartwork:
I think the key is `Unable to connect to ppa.launchpad.net:http:`. Probably a network problem. Looks pretty green now, nice!

hartwork:
Merged, thank you! By any chance, could you give branch `replace-autotools-by-cmake` a spin on Haiku to see if my port to CMake does what it should?

Begasus:
I was looking into it earlier, but couldn't checkout the branch (yet)

hartwork:
One way would be:
```console
# cd "$(mktemp -d)"
# git clone --branch replace-autotools-by-cmake https://github.com/uriparser/uriparser.git
# cd uriparser/
```

PR #46, opened 2019-01-08
build : Add CMakeLists.txt to ease cmake integration

madmongo1:
CMake is a very popular build manager in the c++ world. Including support for it ought to aid adoption.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,32 @@
+cmake_minimum_required(VERSION 3.10)
+
+project(UriParser VERSION 0.9.1 LANGUAGES C)
+
+file(GLOB INTERFACE_FILES
+        RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" CONFIGURE_DEPENDS
+        "include/uriparser/*.h")
+
+file(GLOB SOURCE_FILES
+        RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" CONFIGURE_DEPENDS
+        "src/*.[c|h]")
+
+add_library(uriparser ${INTERFACE_FILES} ${SOURCE_FILES})
+set_property(TARGET uriparser PROPERTY PUBLIC_HEADER "${INTERFACE_FILES}")
+target_include_directories(uriparser PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
+target_include_directories(uriparser PUBLIC $<INSTALL_INTERFACE:include>)
+
+install(TARGETS uriparser
+        EXPORT uriparser
+        LIBRARY
+            DESTINATION lib
+        ARCHIVE
+           DESTINATION lib
+        PUBLIC_HEADER
+            DESTINATION include/uriparser
+        )
+
+install(EXPORT uriparser
+        DESTINATION lib/cmake/uriparser
+        NAMESPACE UriParser::
+        )
+
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,6 @@ Makefile.in
 /uriparser-*.tar.bz2
 /uriparser-*.zip
 /uriparse
+cmake-build-*/
+build*/
+.idea/
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,6 +1,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<project version="4">
-  <component name="VcsDirectoryMappings">
-    <mapping directory="$PROJECT_DIR$" vcs="Git" />
-  </component>
-</project>
--- a/.idea/vcs.xml
+++ b/.idea/vcs.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="$PROJECT_DIR$" vcs="Git" />
+  </component>
+</project>


CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        "include/uriparser/*.h")
+
+file(GLOB SOURCE_FILES
+        RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" CONFIGURE_DEPENDS
+        "src/*.[c|h]")
+
+add_library(uriparser ${INTERFACE_FILES} ${SOURCE_FILES})
+set_property(TARGET uriparser PROPERTY PUBLIC_HEADER "${INTERFACE_FILES}")
+target_include_directories(uriparser PUBLIC $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>)
+target_include_directories(uriparser PUBLIC $<INSTALL_INTERFACE:include>)
```
hartwork: Do you have a good pointer on where to learn more about `$<BUILD_INTERFACE:...>` and `$<INSTALL_INTERFACE:...>`?

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+
+project(UriParser VERSION 0.9.1 LANGUAGES C)
+
+file(GLOB INTERFACE_FILES
+        RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" CONFIGURE_DEPENDS
+        "include/uriparser/*.h")
+
+file(GLOB SOURCE_FILES
+        RELATIVE "${CMAKE_CURRENT_SOURCE_DIR}" CONFIGURE_DEPENDS
+        "src/*.[c|h]")
```
hartwork: FYI I found this warning at https://cmake.org/cmake/help/v3.0/command/file.html:
> (We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.)

CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
...
+        ARCHIVE
+           DESTINATION lib
+        PUBLIC_HEADER
+            DESTINATION include/uriparser
+        )
+
+install(EXPORT uriparser
+        DESTINATION lib/cmake/uriparser
+        NAMESPACE UriParser::
+        )
```
hartwork: Could you elaborate what this last section is for?

CMakeLists.txt
madmongo1: These are called "Generator Expressions". We use a generator here because we want the target to have different public include paths when installed as a package (INSTALL) and when being built as a target of a larger CMAKE build tree (e.g. add_subdirectories or when running tests).

https://cmake.org/cmake/help/v3.13/manual/cmake-generator-expressions.7.html

CMakeLists.txt
madmongo1: This warning is valid unless you use the new "CONFIGURE_DEPENDS" option which supersedes it.
However, I do in fact tend to name each source file when building larger projects. These days it's become a matter of taste.
Globs can make the build system DRYer but do run the risk of including a file you didn't intend to be there.

CMakeLists.txt
madmongo1: When you call `find_package(XXX)` in a cmakelists file, cmake will look in a number of places, first it will look for a module called `FindXXX.cmake` in the `CMAKE_MODULE_PATH`. If that's not found, it will search the `CMAKE_PREFIX_PATH` to search for the files that this line creates. In the following example I ran cmake with `-DCMAKE_INSTALL_PREFIX=${HOME}/install/native-cxx17/uriparser/Release`
The output goes here:
```
$ ls -la lib/cmake/uriparser/
total 16
drwxr-xr-x  4 richardhodges  richardhodges   128 Jan  7 13:06 .
drwxr-xr-x  3 richardhodges  richardhodges    96 Jan  7 12:20 ..
-rw-r--r--  1 richardhodges  richardhodges   865 Jan  7 13:06 uriparser-release.cmake
-rw-r--r--  1 richardhodges  richardhodges  3271 Jan  7 11:47 uriparser.cmake
```

and

```
$ ls -la lib/
total 264
drwxr-xr-x  4 richardhodges  richardhodges     128 Jan  7 13:06 .
drwxr-xr-x  4 richardhodges  richardhodges     128 Jan  7 13:06 ..
drwxr-xr-x  3 richardhodges  richardhodges      96 Jan  7 12:20 cmake
-rw-r--r--  1 richardhodges  richardhodges  132776 Jan  7 13:06 liburiparser.a
```

This means that anyone building this package can now simply install into their own directory, set URIPARSER_DIR to `{prefix}/lib/cmake/uriparser` and then
```
find_package(uriparser CONFIG REQUIRED)
```
The URIPARSER_DIR step is unnecessary if it's been installed into a standard place (e.g. /ur/local/)

This will mean that the script importing uriparser can simply:
```
find_package(uriparser CONFIG REQUIRED)
target_link_libraries(my-target PUBLIC UriParser::uriparser) # namespace::target
```

And all include/library dependencies will be both set up for this target and propagated to any other target that uses `my-target`.

madmongo1: In the project for which I have used uriparser, I actually use `ExternalProject_Add` for now. One of the problems with that is that you don't get the luxury of being able to use the packaging system (because the targets resulting from ExternalProject_Add are built at build time and cmake's find_package needs to resolve packages at configuration time).

In this case I simply build a mock target - which is essentially what the find_package code is doing:

```
ExternalProject_Add(EP_URIPARSER
        PREFIX "${CMAKE_CURRENT_BINARY_DIR}"
        GIT_REPOSITORY "https://github.com/QuoineFinancial/uriparser.git"
        GIT_TAG "master"
        GIT_PROGRESS TRUE
        CMAKE_ARGS
            -DCMAKE_TOOLCHAIN_FILE:FILEPATH=${CMAKE_TOOLCHAIN_FILE}
            -DCMAKE_INSTALL_PREFIX:PATH=<INSTALL_DIR>
            -DCMAKE_BUILD_TYPE:STRING=Release
        )
ExternalProject_Get_Property(EP_URIPARSER INSTALL_DIR)
set(uriparser_install_dir "${INSTALL_DIR}")
add_library(uriparser-dependency INTERFACE)
add_dependencies(uriparser-dependency EP_URIPARSER)
target_include_directories(uriparser-dependency SYSTEM INTERFACE "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>")
target_link_libraries(uriparser-dependency INTERFACE "$<BUILD_INTERFACE:uriparser>")
target_link_directories(uriparser-dependency INTERFACE "$<BUILD_INTERFACE:${uriparser_install_dir}/lib>")
add_library(UriParser::uriparser ALIAS uriparser-dependency)

```

CMakeLists.txt
hartwork: Interesting!
It turns out that `CONFIGURE_DEPENDS` does not exist prior to 3.12 which is more recent than Ubuntu xenial, which would be at least some problem, e.g. regarding CI. I'll keep it in mind, Let's go with a whitelist for now.

CMakeLists.txt
hartwork: I don't get all of it yet but many thanks for the explanation! I have added some export bytes to #47 now.

CMakeLists.txt
hartwork: Thanks!

hartwork:
Superseded by #47, closing this one over here.

PR #49, opened 2019-01-11
Fix VS2005 Build, UriMemory.c, UriMemory.h missing from project

pix64:
Uriparser will compile w/o error, but the linker will throw unresolved external symbol errors when you go to actually use the library.


hartwork:
Hi Adam,

thanks for the report and patch, looks good!

Issue #39, opened 2018-10-27
Make CI cover MinGW cross-compilation


hartwork:
At either one of
- a) [`.travis.yml`](https://github.com/uriparser/uriparser/blob/master/.travis.yml)
- b) [`appveyor.yml`](https://github.com/uriparser/uriparser/blob/master/appveyor.yml)

Nice and clean pull request welcome! :smiley:

KangLin:
See #23

PR #53, opened 2019-01-12
Declare inet_ntop depending on _WIN32_WINNT

jibsen:
So, it seems the reason MSVC was complaining while mingw-w64 was not, is that mingw-w64 defaults to an older SDK version which means ws2tcpip.h does not include the declaration of inet_ntop.

This sets the target version to Vista (required version for [inet_ntop](https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-inet_ntop#requirements)).

### DIFF
--- a/tool/uriparse.c
+++ b/tool/uriparse.c
@@ -36,6 +36,7 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
 #include <stdio.h>
 #include <stdlib.h>
 #include <uriparser/Uri.h>
@@ -43,8 +44,10 @@
 #ifdef _WIN32
 # include <winsock2.h>
 # include <ws2tcpip.h>
-WINSOCK_API_LINKAGE const char WSAAPI inet_ntop(
-		int af, const void *src, char *dst, socklen_t size);
+# if _WIN32_WINNT < 0x0600
+WINSOCK_API_LINKAGE LPCSTR WSAAPI inet_ntop(
+		INT Family, PVOID pAddr, LPSTR pStringBuf, size_t StringBufSize);
+# endif
 #else
 # include <sys/socket.h>
 # include <arpa/inet.h>


tool/uriparse.c
```diff
@@ -36,15 +36,21 @@
  * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  */
+
+#ifdef _WIN32
+# ifndef _WIN32_WINNT
+#  define _WIN32_WINNT 0x0600
+# endif
+# include <SDKDDKVer.h>
```
hartwork: Does `SDKDDKVer.h` exist with MinGW?

jibsen:
I tested this with mingw-w64, MSYS2, Cygwin, and MSVC, and it worked with all of them. But now I installed the original MinGW, and it does not include inet_ntop in ws2tcpip.h. This is going to take a little more experimenting.

jibsen:
Some random notes while looking into this:

Since inet_ntop is Vista+, I believe the "[right way](https://docs.microsoft.com/da-dk/windows/desktop/WinProg/using-the-windows-headers)" would be to define _WIN32_WINNT (and possibly NTDDI_VERSION). But that doesn't help if we need to support compilers that do not have up-to-date headers (original MinGW).

According to [this issue](https://sourceforge.net/p/mingw-w64/bugs/764/), the declaration of inet_ntop in the MSVC headers changed in the Windows 10 SDK so the second parameter is a const pointer. This means we cannot simply declare the function unconditionally.

We could guard the declaration to only be included when the compiler is not MSVC, but this could break if any of the other compilers chooses to update their default SDK version to Vista or above, which brings in the declaration in their headers.

I am leaning towards declaring inet_ntop depending on the value of _WIN32_WINNT.

hartwork:
I think we need MinGW support in the CI first (#39).

jibsen:
I can see you added a check for MinGW, as mentioned above this will work as long as MinGW does not update their default SDK version, so closing this.

hartwork:
@jibsen pull request #131 may be of interest. It adds the `_WIN32_WINNT < 0x0600` bits in its current form.

PR #54, opened 2019-01-15
Fix include of internal C header in C++ test

jibsen:
There are multiple ways to fix this:

  - Add (guarded) extern "C" to all internal include files in src
  - Declare just the variable we need in the C++ file instead of including the internal header
  - Add extern "C" around the include in the C++ file

I would probably go with the first, but I wasn't sure to which extent it might affect the double inclusion trickery. The second option duplicates the declaration, which is not good. So I went with option three, which doesn't change the existing code.

Btw, sorry about the two unrelated commits, it just seemed silly to open a separate PR for them (and they are at least header-related).

### DIFF
--- a/src/UriMemory.h
+++ b/src/UriMemory.h
@@ -37,6 +37,11 @@
  * OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
+#ifndef URI_MEMORY_H
+#define URI_MEMORY_H 1
+
+
+
 #ifndef URI_DOXYGEN
 # include <uriparser/Uri.h>
 #endif
@@ -57,3 +62,7 @@
 extern UriMemoryManager defaultMemoryManager;
 
 UriBool uriMemoryManagerIsComplete(const UriMemoryManager * memory);
+
+
+
+#endif /* URI_MEMORY_H */
--- a/test/MemoryManagerSuite.cpp
+++ b/test/MemoryManagerSuite.cpp
@@ -25,7 +25,11 @@
 #include <gtest/gtest.h>
 
 #include <uriparser/Uri.h>
+
+// For defaultMemoryManager
+extern "C" {
 #include "../src/UriMemory.h"
+}
 
 
 namespace {
--- a/test/MemoryManagerSuite.cpp
+++ b/test/MemoryManagerSuite.cpp
@@ -19,9 +19,9 @@
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
  */
 
-#include <cassert>
-#include <cerrno>
-#include <cstring>  // memcpy
+#include <assert.h>
+#include <errno.h>
+#include <string.h>  // memcpy
 #include <gtest/gtest.h>
 
 #include <uriparser/Uri.h>


jibsen:
As mentioned, my personal preference is also to add `extern "C"` to all C header files because somebody might try to compile the C code with a C++ compiler at some point. Would you prefer it only on the one header file or all of them?

> Any ideas how to make my G++/Clang++ fail in my face for the current lack of extern "C"?

As I understand it, not mangling global variable names is part of the ABI GCC uses (and Clang when not targeting MSVC), and I don't know of any way to modify this (outside perhaps linker scripts).

I think a better option would be to set up a Windows CI to build with MSVC (which should be fairly easy now we have CMake).

> For the change from <cassert> to <assert.h> (and friends) I would rather forward with using namespace std; or by prepending explicit std:: where needed, instead. How do these three approaches compare to you?

I would advice against `using namespace std;`. It's hard to tell in what interesting ways it might break your code on any given platform (and in a future C++ standard). I think it should really only be used for slides and small examples where the `std::` get in the way of the message.

The C++ standard says that `<cfoo>` will add its contents to namespace std, and may add it to the global namespace, whereas `<foo.h>` will add it to the global namespace, and may add it to namespace std. With the compilers I usually use, `<cfoo>` also adds to the global namespace, but depending on this is not portable.

Personally, I am fine with either choice of including `cfoo` and adding `std::`, or including `foo.h` and using the global names. What is your preference?

> I'm happy to do the hunt for places missing std:: if I have some way to make my G++/Clang++ fail for status quo.

I believe both GCC and MSVC have chosen to make `<foo.h>` only add to the global namespace, and `<cfoo>` include `foo.h` and add aliases for these global names in namespace std. This means if you include `<cfoo>` with these compilers, you will have both and get no errors (like now). But as mentioned, you probably should not rely on this being the case.

hartwork:
> As mentioned, my personal preference is also to add extern "C" to all C header files because somebody might try to compile the C code with a C++ compiler at some point. Would you prefer it only on the one header file or all of them?

If all C code is built as C++ then `extern "C"` would not be needed. To my understanding we need `extern "C"` for mixed linking of C with C++ and only then.

I found that
* `Uri.h` already [contains `extern "C"`](https://github.com/uriparser/uriparser/blob/f8ccdfd76b2d46b7162c17e01baee4de402947b5/include/uriparser/Uri.h#L78).
* `UriBase.h` is included by `Uri.h`
* `UriIp4.h` was lacking `extern "C"` but I have added it on `master` by now. It currently expected to be included in addition to `Uri.h` where needed

While working on the AppVeyor CI today I went with 69066afe41fcba85a8de90ed49c951365d2b8119 for `defaultMemoryManager` for a start to not be blocked. Does that approach look okay to you?

Other headers don't need to be touched any more, if I'm not mistaken.

> Personally, I am fine with either choice of including cfoo and adding std::, or including foo.h and using the global names. What is your preference?

My understanding is that `foo.h` is "C past" and "not really C++" so I'd want to avoid that option. I'm okay with (`::foo` and) `std::foo` if that is what it takes.

> I think a better option would be to set up a Windows CI to build with MSVC (which should be fairly easy now we have CMake).

I have set up a minimal version of a CMake-based AppVeyor CI on `master` now.

jibsen:
> While working on the AppVeyor CI today I went with 69066af for defaultMemoryManager for a start to not be blocked. Does that approach look okay to you?

As long as defaultMemoryManager is the only thing from that header you depend on, sure.

PR #56, opened 2019-03-01
Stop building uriparse when URIPARSER_BUILD_TOOLS is OFF

kou:


hartwork:
Thanks for your contribution!

PR #59, opened 2019-04-09
cmake: Fix dll install path on Win32

myd7349:
Currently, the shared lib build on Win32 doesn't work properly:
1. The generated DLL file exports no symbols, so a corresponding export lib will not be created. And the user of this lib has nothing to link to. This issue can be resolved by setting `CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS`: https://cmake.org/cmake/help/latest/prop_tgt/WINDOWS_EXPORT_ALL_SYMBOLS.html
2. The generated DLL file installed to a wrong path. It is similar to: https://stackoverflow.com/questions/22278381/cmake-add-library-followed-by-install-library-destination

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -169,6 +169,9 @@ install(
     LIBRARY
         DESTINATION
             ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME
+        DESTINATION
+            ${CMAKE_INSTALL_BINDIR}
     PUBLIC_HEADER
         DESTINATION
             ${CMAKE_INSTALL_INCLUDEDIR}/uriparser


CMakeLists.txt
```diff
@@ -169,6 +170,9 @@ install(
     LIBRARY
         DESTINATION
             ${CMAKE_INSTALL_LIBDIR}
+    RUNTIME
+        DESTINATION
+            ${CMAKE_INSTALL_BINDIR}
```
hartwork: Why `CMAKE_INSTALL_BINDIR` rather than `CMAKE_INSTALL_LIBDIR`?

CMakeLists.txt
```diff
@@ -62,6 +62,7 @@ option(URIPARSER_BUILD_TOOLS "Build tools (e.g. CLI \"uriparse\")" ON)
 option(URIPARSER_BUILD_CHAR "Build code supporting data type 'char'" ON)
 option(URIPARSER_BUILD_WCHAR_T "Build code supporting data type 'wchar_t'" ON)
 set(URIPARSER_MSVC_RUNTIME "" CACHE STRING "Use of specific runtime library (/MT /MTd /MD /MDd) with MSVC")
+set(CMAKE_WINDOWS_EXPORT_ALL_SYMBOLS ON)
```
hartwork: uriparser has a pile of non-public symbols that would be exposed this way, right?
Is this needed for MSVC only? Is selective use of `__declspec(dllexport)` and `__declspec(dllimport)` the alternative approach here? My understanding is that during library compilation we'd need `__declspec(dllexport)` while during application compilation we'd need `__declspec(dllimport)` for the same symbols. I guess `__declspec(dllexport)` would best go to the related `.c` files while `__declspec(dllimport)` would go to .h files? Does that sounds about right? I could make a pull request with that selective approach for you to review. What do you think?

CMakeLists.txt
myd7349: According the CMake docs(https://cmake.org/cmake/help/latest/command/install.html):
>Executables are treated as RUNTIME objects, except those marked with the MACOSX_BUNDLE property on macOS (see BUNDLE below.) For DLL platforms (all Windows-based systems including Cygwin), the DLL part of a shared library is treated as a RUNTIME target.

Since executables(like `uriparse`) are installed to `bin`, there is no reason to install the DLLs to lib directory. Otherwise, executables will not find the DLLs they relay on.

Installing the `RUNTIME` to `CMAKE_INSTALL_BINDIR` is taken by many widely used projects:
https://github.com/edenhill/librdkafka/blob/master/src/CMakeLists.txt#L288
https://github.com/QuantStack/xeus/blob/master/CMakeLists.txt#L312
https://github.com/zeromq/libzmq/blob/master/CMakeLists.txt#L1367

CMakeLists.txt
myd7349: Sure.

hartwork: That PR has come to life at #60 now

CMakeLists.txt
myd7349: @hartwork Sure. Thanks!

hartwork: Marged, thank you!

PR #69, opened 2019-05-01
CMakeLists.txt: fix build without C++

ffontaine:
Specify that uriparser is a C project (C++ is needed only for the test
runner) otherwise build will fail if no C++ compiler is found by cmake

Fixes:
 - http://autobuild.buildroot.org/results/1e191676f28905a81de6282e07978aa5d4f02039

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,8 @@ cmake_minimum_required(VERSION 3.3)
 project(uriparser
     VERSION
         0.9.3
+    LANGUAGES
+        C
 )
 
 # See https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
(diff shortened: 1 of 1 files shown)


hartwork:
Hi @ffontaine , thanks for taking this upstream!

What concerns me a bit right now is that the proposed change affects all targets — including the C++ testrunner target — rather than just the library and commend line tool. Is there a way to make the list of languages depend on `URIPARSER_BUILD_TESTS` or to set it per target rather than for the whole project?

Thanks!

arnout:
To enable C++ for the testrunner target you can use the [ENABLE_LANGUAGE](https://cmake.org/cmake/help/v3.0/command/enable_language.html) command within the condition that enables the testrunner.

ffontaine:
Thanks @arnout, I'll test it and update my PR.

hartwork:
@ffontaine I found a way now to get AppVeyor CI happy. The problem was the CXX build flags were modified before CXX support was enabled (since it's happening a lot later now). cb0001dc6d53027ec0cde04f76bb0fc6bca1a3af addresses that which allowed to merge your changes into `master` now. Thanks for your contribution to uriparser!

PR #72, opened 2019-07-11
print the configuration at status level

codeInnovation:
When printing the configuration via the CMake message command, it is helpful to specify the type of message.  When no type is specified then the message will print to stderr.  The information printed seems to be status related.  For some build systems, printing to stderr will break the build.


hartwork:
Hi @codeInnovation , sure — looks good!

Can you share some details about the environment that turns stderr content to failure? Some if that might be helpful for a future change log entry about it to explain this change better. Thanks!

Best, Sebastian

codeInnovation:
Sure, typically it's build scripts written in bash or alternatives that catch stderr and turn it into an error.

hartwork:
Okay, I was expecting to maybe learn that *&lt;big CI tool XYZ&gt;* would be the case. Okay, thanks!

Issue #74, opened 2019-08-02
Need a way to disable installation of header files


codeInnovation:
I need a way to disable the installation of the files such as the public header files when built with CMake.  If you try to embed the uriparser as a static library, these files get automatically installed with the package.  When two installers both try to include the uriparser library, then there is a collision and it becomes impossible to install both packages.  In other words, I'm looking for an option like Google Test's INSTALL_GTEST in the uriparser library, that can be switched off when needed.

What are your thoughts on this?  Is there a way to turn off installation of files today?  I didn't see a way yet in the CMakeLists.txt file.  I'm certainly willing to help make the change.

hartwork:
In other words you need a switch to bypass all of https://github.com/uriparser/uriparser/blob/ba9990a3e7cbdc073e55aeac6d5e04379248ee3b/CMakeLists.txt#L187-L204 ?

Are you aware of the problems described at https://wiki.gentoo.org/wiki/Why_not_bundle_dependencies ?

codeInnovation:
Yes I was hoping for a switch to disable the install command.  There are several calls to install in CMakeLists.txt.  I have a change on a fork that bypasses them all.  I wasn't really certain if they all needed to be bypassed though or if just the one you have shown would be sufficient.

As for the problems with static linking, I guess that is true.  It does have drawbacks but deployment environments can be challenging sometimes.

hartwork:
Do you want to share your version as a pull request or would you rather want to review a pull request done by me? For a name I'd like `URIPARSER_INSTALL`, I think. Default would be `ON`.

codeInnovation:
I'll submit a pull request.  Thanks!

hartwork:
Any news?

PR #75, opened 2019-08-02
Add option to enable/disable installation of files, default is enabled (#74)

codeInnovation:
for use when bundling uriparser as a static lib, provide a way to optionally turn off installation of files.  the default is behavior of the option is set to on


hartwork:
Related to #74

codeInnovation:
That sounds good to me.  I'll make the change.  For the name of the macro, I plan on following the convention where the prefix of the library is used (to prevent name conflicts with other libraries).  That would fit with the other macro I see called `uriparser_apply_msvc_runtime_to`, so the new one would be named `uriparser_install`.

codeInnovation:
I renamed the option to `URIPARSER_ENABLE_INSTALL`  otherwise it reads a little odd to me having both an option and macro that only differ by upper and lowercase.

codeInnovation:
Sorry about that!  I've changed the white space to spaces instead of tabs.

PR #82, opened 2019-09-05
Fixed license definition

sergeyklay:
Hello,

I fixed BSD 3-Clause License definition.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -17,7 +17,7 @@ are met:
       disclaimer   in  the  documentation  and/or  other  materials
       provided with the distribution.
 
-    * Neither  the name of the <ORGANIZATION> nor the names of  its
+    * Neither the name of the copyright holder nor the names of its
       contributors  may  be  used to endorse  or  promote  products
       derived  from  this software without specific  prior  written
       permission.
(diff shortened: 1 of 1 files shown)


hartwork:
I get the idea of the first half but what's the idea with the second half?
Since this is legal text, to be honest I'm not sure yet I want to touch it all.

sergeyklay:
@hartwork In fact there is a canonical form of the BSD 3-Clause license :)

- https://spdx.org/licenses/BSD-3-Clause.html
- https://opensource.org/licenses/BSD-3-Clause
- https://choosealicense.com/licenses/bsd-3-clause

hartwork:
Hello again,

sorry it took me so long to act on this. I still wonder when that change from owner to holder in the text happened and how the mix in here came to be, precisely. Anyway, I have mass-applied a version of this change with [justified test](http://www.nicemice.net/par/) in #86 now. What do you think?

Best, Sebastian

sergeyklay:
Looks good to me

hartwork:
Thanks for the review and for bringing this up!

PR #89, opened 2020-05-31
@section autoconf Autoconf Check

crawlserv:
old version number in text, but not in autoconf code


hartwork:
Good catch — thank you! :+1:

PR #90, opened 2020-06-01
cmake: fix static lib name for MinGW

SpaceIm:
https://github.com/uriparser/uriparser/pull/64 changed not only dll name but also name of static lib, which lead to wrong extension.
This PR fixes extension for static lib with MinGW, while keeping version suffix in dll name.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,10 @@ if(NOT MSVC)
     set_property(TARGET uriparser PROPERTY VERSION ${URIPARSER_SO_CURRENT_MINUS_AGE}.${URIPARSER_SO_AGE}.${URIPARSER_SO_REVISION})
     set_property(TARGET uriparser PROPERTY SOVERSION ${URIPARSER_SO_CURRENT_MINUS_AGE})
     if(WIN32)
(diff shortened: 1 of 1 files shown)


hartwork:
Hi @SpaceIm , thanks for bringing this up. Could you elaborate on the part about the wrong extension? E.g. what is wrong, what is right, to whom, and why?

@manisandro what do you think about this change?

SpaceIm:
@hartwork Sure I can elaborate ;)

Basically with MinGW:
- if building shared uriparser, it creates those files: `liburiparser.dll.a` and `liburiparser-1.dll`
- if building static uriparser, it creates this file archive file: `liburiparser-1.dll` => Wrong, file name should be `liburiparser.a` (or `liburiparser.lib`). This PR addresses this issue.

hartwork:
So we cannot use `${CMAKE_SHARED_LIBRARY_SUFFIX}` without checking if we're building a shared library, I guess. Makes sense.

Why is there no version suffix with the static `.(a|lib)` files, e.g. why not `liburiparser-1.(a|lib)`?

SpaceIm:
Why would you want version suffix in static lib name on windows (or other os)? It's uncommon.

hartwork:
Maybe you're right. Let's see what @manisandro thinks so we don't break things for him.

manisandro:
Change looks good to me - correct, you'll not want the version suffix for the static library.

hartwork:
Excellent, thanks to both of you! :+1: :+1:

PR #91, opened 2020-09-25
pkg-config: improve template to be more generic

Cynerd:
This also removes issues when cross-compiling for example in OpenWrt.

### DIFF
--- a/liburiparser.pc.in
+++ b/liburiparser.pc.in
@@ -1,7 +1,7 @@
 prefix=@CMAKE_INSTALL_PREFIX@
-exec_prefix=@CMAKE_INSTALL_PREFIX@
-libdir=@CMAKE_INSTALL_FULL_LIBDIR@
-includedir=@CMAKE_INSTALL_FULL_INCLUDEDIR@
+exec_prefix=${prefix}
+libdir=${exec_prefix}/@CMAKE_INSTALL_LIBDIR@
+includedir=${prefix}/@CMAKE_INSTALL_INCLUDEDIR@
 
 Name: liburiparser
 Description: URI parsing and handling library


hartwork:
Please provide more details about what the problem is and why this particular fix seems right to you. I'm not saying that something is wrong, I just cannot blind merge things without a proper bug report and full understanding. Do you have a link on where this caused trouble for OpenWrt in practice, for instance? Thank you!

Cynerd:
Variables `prefix` and `exec_prefix` are common and expected variables and are used in case of OpenWrt to dynamically set prefix. The point is that build system such as OpenWrt's can be moved on filesystem and inserting absolute path is invalid. Build system knows where it is and where the root is so it injects appropriate `prefix` and `exec_prefix` (also note that those can be different). Other option would be to either always hack pc file with some sed or rebuild all packages when build directory is moved.

Real code taking advantage of this is here: https://github.com/openwrt/openwrt/blob/master/tools/pkgconf/files/pkg-config

hartwork:
@Cynerd I have checked https://github.com/openwrt/openwrt/blob/master/tools/pkgconf/files/pkg-config now; it helped to understand, thank you! Just to be sure: Are there reasons why OpenWRT cannot use `--define-variable` to override `includedir` and `libdir` in the same generic fashion, as well?

Cynerd:
Well I am not sure about that.

My humble opinion is: that might break other packages. It makes sense that package does not have to conform to standard `includedir` and `libdir` paths, such as if that is some sort of plugin system. Overriding every single possible (although standard) variable just could create even bigger problems.

I initially investigated this and looked in to examples for pkg-config files in cmake and most of them used what you have right now. On the other hand if you take a look in to `man 5 pc` there is example:
```
     prefix=/home/kaniini/pkg
     exec_prefix=${prefix}
     libdir=${exec_prefix}/lib
     includedir=${prefix}/include
```
This format can exact combination of variables can be also found in any documentation regarding pkg-config itself. My opinion is that someone on the Internet just originally created example of how to generate `pc` file using cmake and everyone copied it even when it wasn't fully pkg-config advised format.

hartwork (approved):
> My humble opinion is: that might break other packages.

I can now confirm, yes it would break things:

```console
# grep -h '^includedir=' /usr/lib/pkgconfig/*.pc | sort -u | wc -l
41
```

With some digging, I also found related commit https://github.com/openwrt/openwrt/commit/71c8fa813537012ec912415f3a1c52b27428ae86 in OpenWRT.

Alright, I think your pull request is on the right track.

hartwork:
Merged now, thanks for taking this upstream and helping out with a pull request! :pray:

PR #96, opened 2021-02-25
.travis.yml: Enable UBSan with Clang as well

hartwork:


hannob:
For this to practically be useful you need to use ubsan in a way that will not just report errors, but terminate. You should add -fno-sanitize-recover=all for this.

Leak sanitizer is implicitly enabled by address sanitizer, but it does no harm to explicitly enable it.

hartwork:
> For this to practically be useful you need to use ubsan in a way that will not just report errors, but terminate. You should add -fno-sanitize-recover=all for this.

That is **great** to know! I just confirmed with libexpat prior to https://github.com/libexpat/libexpat/pull/438 :
```console
# ./tests/runtests ; echo $?
Expat version: expat_2.2.10
[..]/expat/lib/xmlparse.c:1897:23: runtime error: applying zero offset to null pointer
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior [..]/expat/lib/xmlparse.c:1897:23 in 
100%: Checks: 332, Failed: 0
0
```

PR #97, opened 2021-02-26
Fix RemoveDotSegmentsEx for URIs like "http://a/.." (fixes #92)

### LINKED ISSUE #92: uriAddBaseUri seen omitting trailing slash
Hi - I came across some failing tests in code of mine that uses uriparser, so I reduced them to tests I could add to uriparser's test/test.cpp. I'm probably just not understanding what should be happening according to the spec, but I thought I would ask here.

At lines 1067 and 1068, https://github.com/uriparser/uriparser/blob/1bd8ae2b8b47d3155765c1ea282a8f9812d3853e/test/test.cpp#L1067-L1068 you see two tests from the spec where the relative uris "../.." and "../../" are added to the base uri. In both cases, the resolved uri should have a final "/", whether or not the relative uri does.

The following test statements I wrote show that if you add a "." segment within the base uri, then the relative uri still has a final "/" in both cases:
```
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/c/./d;p?q", L"../..", L"http://a/"));
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/c/./d;p?q", L"../../", L"http://a/"));
```
However, the following statements show that if you add a ".." segment within the base uri, then the resolved uri seems to be missing the final "/" if the relative uri does:
```
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/c/../d;p?q", L"../..", L"http://a/")); // FAILS
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/c/../d;p?q", L"../../", L"http://a/"));
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/../c/d;p?q", L"../..", L"http://a/")); // FAILS
    ASSERT_TRUE(testAddBaseHelper(L"http://a/b/../c/d;p?q", L"../../", L"http://a/"));
```
Also note that if the ".." segment is close enough to the authority segment, then the final "/" reappears:
```
    ASSERT_TRUE(testAddBaseHelper(L"http://a/../b/c/d;p?q", L"../..", L"http://a/")); // SUCCEEDS
    ASSERT_TRUE(testAddBaseHelper(L"http://a/../b/c/d;p?q", L"../../", L"http://a/"));
```
Am I just doing something wrong here?
thanks

hartwork:
Fixes #92

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -4,6 +4,10 @@ NOTE: uriparser is looking for help with a few things:
 
 20xx-xx-x -- x.x.x
 
+  * Fixed: Fix a bug regarding section "5.2.4. Remove Dot Segments"
+      of RFC 3986 that affected both normalization and reference resolution
+      with regard to trailing slashes.
+      Thanks to Dan Pape for the report!
   * Fixed: MinGW: Fix name of static library (GitHub #90)
       Thanks to SpaceIm for the patch and Sandro Mani for review!
   * Fixed: Use correct inline marker "__forceinline" for Intel C++ Compiler
--- a/ChangeLog
+++ b/ChangeLog
@@ -6,7 +6,7 @@ NOTE: uriparser is looking for help with a few things:
 
   * Fixed: Fix a bug regarding section "5.2.4. Remove Dot Segments"
       of RFC 3986 that affected both normalization and reference resolution
-      with regard to trailing slashes.
+      with regard to trailing slashes (GitHub #92)
       Thanks to Dan Pape for the report!
   * Fixed: MinGW: Fix name of static library (GitHub #90)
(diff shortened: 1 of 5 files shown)


danpape:
Your changes all make sense to me and using this version of the code, all of my local tests now pass. I really like how you made your changes in a few commits (rather than just one like I normally would) as that made it much easier to follow your train of thought. Thank you.

hartwork:
@danpape that is great news, thanks for your review and checking! :pray: I'll merge this in then…

PS: If you're interested, I wrote a blog post that touches upon topic "coding with review in mind" at https://blog.hartwork.org/posts/my-approach-to-code-review/ .

PR #98, opened 2021-03-04
Add CMAKE_CURRENT_SOURCE_DIR prefix to sources

shehzan10:
When the prefix is missing, installing `uriparser` from dependent projects causes issues as the files before relative to the parent directories.

This should fix that by using absolute paths from the prefix.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -117,34 +117,34 @@ configure_file(src/UriConfig.h.in config.h)
 # C library
 #
 set(API_HEADER_FILES
-    include/uriparser/UriBase.h
-    include/uriparser/UriDefsAnsi.h
-    include/uriparser/UriDefsConfig.h
-    include/uriparser/UriDefsUnicode.h
-    include/uriparser/Uri.h
-    include/uriparser/UriIp4.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/UriBase.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/UriDefsAnsi.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/UriDefsConfig.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/UriDefsUnicode.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/Uri.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/include/uriparser/UriIp4.h
 )
 set(LIBRARY_CODE_FILES
-    src/UriCommon.c
-    src/UriCommon.h
-    src/UriCompare.c
-    src/UriEscape.c
-    src/UriFile.c
-    src/UriIp4Base.c
-    src/UriIp4Base.h
-    src/UriIp4.c
-    src/UriMemory.c
-    src/UriMemory.h
-    src/UriNormalizeBase.c
-    src/UriNormalizeBase.h
-    src/UriNormalize.c
-    src/UriParseBase.c
-    src/UriParseBase.h
-    src/UriParse.c
-    src/UriQuery.c
-    src/UriRecompose.c
-    src/UriResolve.c
-    src/UriShorten.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCommon.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCommon.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCompare.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriEscape.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriFile.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriIp4Base.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriIp4Base.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriIp4.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriMemory.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriMemory.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriNormalizeBase.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriNormalizeBase.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriNormalize.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriParseBase.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriParseBase.h
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriParse.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriQuery.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriRecompose.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriResolve.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriShorten.c
 )
 
 add_library(uriparser
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -217,7 +217,7 @@ uriparser_install(
 #
 if(URIPARSER_BUILD_TOOLS)
     add_executable(uriparse
-        tool/uriparse.c
+        ${CMAKE_CURRENT_SOURCE_DIR}/tool/uriparse.c
     )
 
     target_link_libraries(uriparse PUBLIC uriparser)
@@ -396,7 +396,7 @@ uriparser_install(
 # pkg-config file
 #
 if(NOT MSVC)
-    configure_file(liburiparser.pc.in liburiparser.pc @ONLY)
+    configure_file(${CMAKE_CURRENT_SOURCE_DIR}/liburiparser.pc.in liburiparser.pc @ONLY)
     uriparser_install(
         FILES
             ${CMAKE_CURRENT_BINARY_DIR}/liburiparser.pc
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
Hi @shehzan10 thanks for this pull request! :+1: I checked the file for more places where this might be needed and found more places. Could you extend the pull request to cover all code files referenced in `CMakeLists.txt`? Thanks! :pray:

shehzan10:
@hartwork Thanks for checking it quickly. If you are referring to line 112 (https://github.com/uriparser/uriparser/pull/98/files#diff-1e7de1ae2d059d21e1dd75d5812d5a34b0222cef273b7c3a2af62eb747f9d20aR112), then that does not need the prefix as that command run at CMake Configuration Time. So they will always be local is my understanding.

Have I missed any others?

hartwork:
PS: @shehzan10 with some Bash I found these 9 referenced files more:
```console
# fgrep -o -f <(git ls-files | grep -v '^configure$') CMakeLists.txt | fgrep -v -e src/ -e include/uriparser/ | tee /dev/stderr | wc -l
tool/uriparse.c
test/FourSuite.cpp
test/MemoryManagerSuite.cpp
test/test.cpp
test/VersionSuite.cpp
doc/Doxyfile.in
doc/release.sh.in
cmake/uriparser-config.cmake.in
liburiparser.pc.in
9
```

hartwork:
I only read https://github.com/uriparser/uriparser/pull/98#issuecomment-790228564 just now. Not line 112, no.

hartwork:
> PS: @shehzan10 with some Bash I found these 9 referenced files more:

Actually 10 I think — second take:
```console
# fgrep -o -f <(git ls-files | grep -v '^configure$') CMakeLists.txt | grep -v -e 'src/.*[ch]$' -e include/uriparser/ | tee /dev/stderr | wc -l
src/UriConfig.h.in
tool/uriparse.c
test/FourSuite.cpp
test/MemoryManagerSuite.cpp
test/test.cpp
test/VersionSuite.cpp
doc/Doxyfile.in
doc/release.sh.in
cmake/uriparser-config.cmake.in
liburiparser.pc.in
10
```

shehzan10:
@hartwork Thanks - I was using the github web editor and it didn't expand far enough.

My personal opinion is that config files shouldn't need the prefix, but based on your request I've added it.

Note that the output is relative to the binary directory and should not use prefix in those cases.

shehzan10:
@hartwork Just wanted to bump this to see if this can be merged, or if you have any other comments.

hartwork:
> @hartwork Just wanted to bump this to see if this can be merged, or if you have any other comments.

@shehzan10 I've been experimenting with this locally now. With regard to [your initial problem description](https://github.com/uriparser/uriparser/pull/98#issue-584417655) I was *assuming* that we are talking about use of uriparser by means of CMake's `add_subdirectory` function. When trying _that_ out though, it already builds and installs without error. :thinking: Before applying patches to any project, in general I try to _really understand them_ and I will also have to write an accurate entry for the change log about what we were fixing here, so I'll need to reproduce the error case prior to the fix some way to be sure I understand the actual problem.

Can you help me with a reproducer? Below is my current Dockerfile that succeeds without error. How would be need to change it to demo the problem and force a build error? Thanks in advance!

```Dockerfile
# Copyright (C) 2021 Sebastian Pipping <sebastian@pipping.org>
# Licensed under Creative Commons CC0 1.0 Universal

FROM debian:sid

RUN apt-get update && apt-get install --yes --no-install-recommends -V \
        build-essential \
        ca-certificates \
        cmake \
        doxygen \
        git \
        graphviz \
        libgtest-dev

SHELL ["bash", "-c"]

RUN mkdir parent \
        && \
    cd parent \
        && \
    git clone --depth 1 https://github.com/uriparser/uriparser uriparser \
        && \
    echo $'find_package(Threads REQUIRED)\ntarget_link_libraries(testrunner PRIVATE Threads::Threads)  # workaround gtest link error' >> uriparser/CMakeLists.txt \
        && \
    echo $'cmake_minimum_required(VERSION 3.18)\nproject(uriapp123)\nadd_subdirectory("uriparser")' > CMakeLists.txt \
        && \
    mkdir build \
        && \
    cd build \
        && \
    cmake .. \
        && \
    make -j2 \
        && \
    make -j2 check \
        && \
    make install DESTDIR="${PWD}"/root VERBOSE=1 \
        && \
    cd root/ \
        && \
    find | sort
```

hartwork:
>     echo $'find_package(Threads REQUIRED)\ntarget_link_libraries(testrunner PRIVATE Threads::Threads)  # workaround gtest link error' >> uriparser/CMakeLists.txt \

FYI with #100 merged, that^^ line should be omitted by now.

shehzan10:
This is the closest way I have been able to replicate it using the set up we are using

```
# Copyright (C) 2021 Sebastian Pipping <sebastian@pipping.org>
# Licensed under Creative Commons CC0 1.0 Universal

FROM debian:sid

RUN apt-get update && apt-get install --yes --no-install-recommends -V \
        build-essential \
        ca-certificates \
        cmake \
        doxygen \
        git \
        graphviz \
        libgtest-dev

SHELL ["bash", "-c"]

RUN mkdir parent \
        && \
    cd parent \
        && \
    mkdir extern \
        && \
    cd extern \
        && \
    git clone --depth 1 https://github.com/uriparser/uriparser uriparser \
        && \
    echo $'option(URIPARSER_BUILD_TESTS "" off)\noption(URIPARSER_BUILD_DOCS "" off)\noption(URIPARSER_ENABLE_INSTALL "" on)\noption(URIPARSER_BUILD_TOOLS "" off)\nadd_subdirectory(uriparser)' > CMakeLists.txt \
        && \
    cd .. \
        && \
    echo $'cmake_minimum_required(VERSION 3.18)\nproject(uriapp123)\noption(BUILD_SHARED_LIBS "Build shared libraries" OFF)\nadd_subdirectory(extern EXCLUDE_FROM_ALL)\ninstall(TARGETS uriparser OPTIONAL)' > CMakeLists.txt \
        && \
    mkdir build \
        && \
    cd build \
        && \
    cmake .. \
        && \
    make -j2 uriparser \
        && \
    make install DESTDIR="${PWD}"/root VERBOSE=1 \
        && \
    cd root/ \
        && \
    find | sort

```

Note: I am explicitly calling `make -j2 uriparser` because there are no other targets depending on uriparser, which would be the case in other projects. Calling `make install uriparser` on the install step will not change the installed result.

hartwork:
Hi @shehzan10 I am not sure I understand. Your Dockerfile uses `add_subdirectory` like mine and succeeds like mine. We need something that fails :smiley:  If you opened this pull request originally because the build failed in your environment and uriparser was the cause for this, can you extract the essence of that failing setup and use that to make a minimal reproducible example that fails? One approach could be to duplicate the setup and keep cutting things away. Would that work?

> Calling `make install uriparser` on the install step will not change the installed result.

Yes, because make understands both arguments `install` and `uriparser` as "targets" and target `uriparser` has are been built successfully before and so for that target make will just produce output `[100%] Built target uriparser` once more.

shehzan10:
@hartwork One thing I missed removing before posting the file above is `OPTIONAL` from

```
    echo $'cmake_minimum_required(VERSION 3.18)\nproject(uriapp123)\noption(BUILD_SHARED_LIBS "Build shared libraries" OFF)\nadd_subdirectory(extern EXCLUDE_FROM_ALL)\ninstall(TARGETS uriparser OPTIONAL)' > CMakeLists.txt \
```

The optional part ignore missing files. Once you remove that, the install step errors with:
```
-- Install configuration: ""
-- Installing: /parent/build/root/usr/local/lib/liburiparser.a
CMake Error at cmake_install.cmake:50 (file):
  file INSTALL cannot find "/parent/include/uriparser/UriBase.h": No such
  file or directory.


make: *** [Makefile:105: install] Error 1
```

As you can see, it is trying to read the file from `/parent/include/uriparser/`, not `/parent/extern/uriparser/include/uriparser`. This is why the `CMAKE_CURRENT_SOURCE_DIR` is useful.

hartwork:
Hi @shehzan10 cool! So _installing_ is key. I was able to reproduce (1) the failure and (2) that `${CMAKE_CURRENT_SOURCE_DIR}` fixes the issue over here. While playing with a mass fix for this — this command…

```
git ls-files | sed 's,.*,-e s|\\\\b\0\\\\b|${CMAKE_CURRENT_SOURCE_DIR}/\0|,' | xargs sed -i CMakeLists.txt
```

…I noticed that the pull request seems to be missing `${CMAKE_CURRENT_SOURCE_DIR}` at two more places still:
```diff
# git --no-pager diff shehzan10-patch-1 
diff --git a/CMakeLists.txt b/CMakeLists.txt
index f11ba63..e427b17 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,7 +226,7 @@ uriparser_install(
 #
 if(URIPARSER_BUILD_TOOLS)
     add_executable(uriparse
-        tool/uriparse.c
+        ${CMAKE_CURRENT_SOURCE_DIR}/tool/uriparse.c
     )
 
     target_link_libraries(uriparse PUBLIC uriparser)
@@ -414,7 +414,7 @@ uriparser_install(
 # pkg-config file
 #
 if(NOT MSVC)
-    configure_file(liburiparser.pc.in liburiparser.pc @ONLY)
+    configure_file(${CMAKE_CURRENT_SOURCE_DIR}/liburiparser.pc.in liburiparser.pc @ONLY)
     uriparser_install(
         FILES
             ${CMAKE_CURRENT_BINARY_DIR}/liburiparser.pc
```
Could you add those two places more?

Thanks and best, Sebastian

shehzan10:
@hartwork I commented about this in https://github.com/uriparser/uriparser/pull/98#issuecomment-790239089.

The output of `configure_file` is relative to the current binary directory, not source directory. So those should not be changed.

https://cmake.org/cmake/help/v3.0/command/configure_file.html
> If <output> is a relative path it is evaluated with respect to the current binary directory.

hartwork:
> The output of `configure_file` is relative to the current binary directory, not source directory. So those should not be changed.

Yes, but the out is the _second_ argument, my point is about adjusting the _first_.

shehzan10:
@hartwork pushed.

hartwork:
Thanks! And thanks for bringing this up and contributing to uriparser! :pray:

PR #103, opened 2021-03-06
CMake: Improve qhelpgenerator auto-detection

hartwork:
Related to https://bugs.gentoo.org/773430


hartwork:
@a17r good enough to merge?

CMakeLists.txt
```diff
@@ -326,11 +326,20 @@ endif()
...
+    set(QHG_LOCATION "" CACHE FILEPATH "Path to qhelpgenerator program (default: auto-detect)")
+    if(NOT QHG_LOCATION)
+        find_package(Qt5Help QUIET)
+        if(TARGET Qt5::qhelpgenerator)
+            get_target_property(QHG_LOCATION Qt5::qhelpgenerator LOCATION)
+            mark_as_advanced(Qt5Core_DIR)
+            mark_as_advanced(Qt5Gui_DIR)
+            mark_as_advanced(Qt5Help_DIR)
+            mark_as_advanced(Qt5Sql_DIR)
+            mark_as_advanced(Qt5Widgets_DIR)
```
a17r, line 339: those lines appear to be unrelated to this change

hartwork: They are actually related because of https://github.com/uriparser/uriparser/blame/master/README.md#L48 . If I don't add these, the readme goes out of sync and these 5 variables show up in that listing while I would rather keep that listing as is.

hartwork: I'm assuming this is good to merge. If that's a misunderstanding, please let me know — we can sure add on top of this in a new pull request…

PR #110, opened 2021-03-23
CMake: Only set CMAKE_CXX_FLAGS after enable_language(CXX)

hartwork:
@arichardson in a sense I'm putting words/commits into your mouth here so it would be awesome if you had a second for review to be sure you don't feel misrepresented after. Thanks!

CC https://github.com/libexpat/libexpat/pull/442#issuecomment-802208332


arichardson:
@hartwork Thanks! this looks good to me.

hartwork:
@arichardson excellent — thank you!

PR #131, opened 2022-01-07
uriparse: use size_t for inet_ntop declaration and fix macro checks for MinGW and mingw-w64

1480c1:
From a patch from msys2, `/usr/x86_64-w64-mingw32/include/_mingw.h` defines `_WIN32_WINNT` on mingw-w64 cross compilers

### DIFF
--- a/tool/uriparse.c
+++ b/tool/uriparse.c
@@ -43,9 +43,11 @@
 #ifdef _WIN32
 # include <winsock2.h>
 # include <ws2tcpip.h>
-# ifdef __MINGW32__
+# if defined(__MINGW32__) &&                                              \
+     (!defined(__MINGW64_VERSION_MAJOR) || __MINGW64_VERSION_MAJOR < 3 || \
+      _WIN32_WINNT < _WIN32_WINNT_VISTA)
 WINSOCK_API_LINKAGE const char * WSAAPI inet_ntop(
-		int af, const void *src, char *dst, socklen_t size);
+		int af, const void *src, char *dst, size_t size);
 # endif
 #else
 # include <sys/socket.h>


tool/uriparse.c
```diff
@@ -43,7 +43,7 @@
 #ifdef _WIN32
 # include <winsock2.h>
 # include <ws2tcpip.h>
-# ifdef __MINGW32__
+# if (_WIN32_WINNT < 0x600)
```
hartwork, line 46: Hi @1480c1 thanks for taking this upstream. What I am wondering right now:
- What is the build error without this patched? Are there build logs somewhere?
- Is use of `_WIN32_WINNT_VISTA` rather than `0x600` an option
- Why drop the MinGW part of the guard?

1480c1: > What is the build error without this patched? Are there build logs somewhere?

```c
[33/38] Building C object CMakeFiles/uriparse.dir/tool/uriparse.c.obj
FAILED: CMakeFiles/uriparse.dir/tool/uriparse.c.obj
C:\msys64\ucrt64\bin\cc.exe -DURI_STATIC_BUILD -IC:/Users/cddeg/uriparser/include -fvisibility=hidden -O3 -DNDEBUG -MD -MT CMakeFiles/uriparse.dir/tool/uriparse.c.obj -MF CMakeFiles\uriparse.dir\tool\uriparse.c.obj.d -o CMakeFiles/uriparse.dir/tool/uriparse.c.obj -c C:/Users/cddeg/uriparser/tool/uriparse.c
C:/Users/cddeg/uriparser/tool/uriparse.c:47:41: error: conflicting types for 'inet_ntop'; have 'const char *(int,  const void *, char *, socklen_t)' {aka 'const char *(int,  const void *, char *, int)'}
   47 | WINSOCK_API_LINKAGE const char * WSAAPI inet_ntop(
      |                                         ^~~~~~~~~
In file included from C:/Users/cddeg/uriparser/tool/uriparse.c:45:
C:/msys64/ucrt64/x86_64-w64-mingw32/include/ws2tcpip.h:401:35: note: previous declaration of 'inet_ntop' with type 'const CHAR *(INT,  const void *, CHAR *, size_t)' {aka 'const char *(int,  const void *, char *, long long unsigned int)'}
  401 | WINSOCK_API_LINKAGE LPCSTR WSAAPI InetNtopA(INT Family, LPCVOID pAddr, LPSTR pStringBuf, size_t StringBufSize);
      |                                   ^~~~~~~~~
```

Here is the build log unpatched, the issue seems to stem from https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_ntop specifying the last argument as `size_t` instead of `socklen_t`

> Is use of `_WIN32_WINNT_VISTA` rather than `0x600` an option

Is it, I can change it to use that one instead, so it's more verbose

> Why drop the MinGW part of the guard?

Not entirely sure on that part, it seems to be a carry-over from https://github.com/msys2/MINGW-packages/blob/b5598a0bf7612941ca53023698bac8e614ddcdcf/mingw-w64-uriparser/uriparser-0.8.6.patch#L56 and the MinGW guard was removed in https://github.com/msys2/MINGW-packages/commit/68e8c85c5fa29beacc1ff78b9da46c20b2a2793d#diff-533561ae756ce35324f9b02c71736528dde66d5184f0e697270ab48274024ca8

I can instead change this part to
```suggestion
# if defined(__MINGW32__) && (_WIN32_WINNT < _WIN32_WINNT_VISTA)
```

to keep the MinGW guard if that's prefered

1480c1: Ah, I think I have a better understanding of why this declaration exists now, inside MinGW (the original)[1], `inet_ntop` only existed inside a .def file with no declaration, but mingw-w64 had `inet_ntop` available as a function declaration since mingw-w64 4.0.1 (if https://github.com/mingw-w64/mingw-w64/commit/175e4bf7e10e42340b54a8092b755bd13006a28f is to be believed).

If this is so, then this might require a different macro check than I initially assumed, since having the `_WIN32_WINNT < _WIN32_WINNT_VISTA` clause would mean a MinGW toolchain targeting windows later than Vista would still lack an inet_ntop declaration. From what I can gather, one macro that exists inside mingw-w64 but not MinGW would be the `__MINGW64_VERSION_MAJOR` version macro, so I can probably use that and change this to either

```suggestion
# ifndef __MINGW64_VERSION_MAJOR
```

or
```suggestion
# if !defined(__MINGW64_VERSION_MAJOR) || (__MINGW64_VERSION_MAJOR < 4)
```

if we want to be more specific


[1] https://osdn.net/projects/mingw/scm/git/mingw-org-wsl/tree/5.4-trunk/w32api/include/

brechtsanders:
I did a similar proposition in https://github.com/uriparser/uriparser/pull/133
Initially I thought checking version numbers as proposed here but in the end I went for:
```
# if defined(__MINGW32__) && !defined(InetNtopA)
```
as that really checks for the actual issue.

1480c1:
The one you have proposed seems reasonable to me as well as in the commit in mingw-w64 that introduced inet_ntop, they introduce that define as well, so it would have a similar effect but would also work on the commits in between mingw-w64 4 and the actual commit that introduced the declaration.

My only potential concern would be if mingw-w64 might ever change that line, but considering how long it's lasted, I don't think that concern will be valid anytime soon. Another concern would be with MinGW and if they will ever introduce a declaration for this function, but I think we can cross that bridge if that situation ever arises

I am in favor of using #133 over the proposed changes in here

tool/uriparse.c
hartwork: Hi again @1480c1, @brechtsanders, @manisandro,

great research and ideas, I appreciate it!  I went verifying them now.  Let me summarize:

- (There is both MinGW (ex MinGW32) and MinGW-w64 (and I assume we want to support both))

- MinGW-w64 introduced `inet_ntop` in commit [`175e4bf7e10e42340b54a8092b755bd13006a28f`](https://github.com/mingw-w64/mingw-w64/commit/175e4bf7e10e42340b54a8092b755bd13006a28f)
  as @1480c1 pointed out *but* that commit is already part of MinGW-w64  *3.0.0* as proven by
  ```
  # git tag --contains 175e4bf7e1 | sed 's/^v//' | sort -V | head -n1
  3.0.0
  ```

- MinGW (ex MinGW32) has header `ws2tcpip.h` in package `w32api-5.0.2`
  and they don't have a prototype for `inet_ntop` in there or anywhere,
  just the linker symbol:
  ```
  # fgrep -R inet_ntop w32api-5.0.2/
  w32api-5.0.2/lib/ws2_32.def:inet_ntop@16
  ```
  @1480c1 found the same.

- Prototypes of `inet_ntop` use `size_t` (rather than `socklen_t`) for the last
  parameter in both [MSDN](https://docs.microsoft.com/en-us/windows/win32/api/ws2tcpip/nf-ws2tcpip-inet_ntop) and MinGW-w64.

- (MinGW-w64  changed `void *` to `const void *` in commit [`0a57d9f2a70f54a9d93d18a6a2505cc67f58f3ca`](https://github.com/mingw-w64/mingw-w64/commit/0a57d9f2a70f54a9d93d18a6a2505cc67f58f3ca) starting release 7.0.0 for the second argument of `inet_ntop`.)

- Proprocessor guard `__MINGW32__` [works](https://sourceforge.net/p/predef/wiki/Compilers/#mingw-and-mingw-w64) for MinGW (ex MinGW32) and MinGW-w64

- MinGW-w64 has `#define InetNtopA inet_ntop` since their first introduction of `inet_ntop` and that explains @brechtsanders's idea in adding `&& !defined(InetNtopA)`. While it can be argued, that that macro is part of the public API of MinGW-w64 now and hence "cannot" be changed or moved, it feels like an implementation detail that I'm not sure is the best to rely on.

- MinGW-w64 has always been guarding `inet_ntop` by `_WIN32_WINNT >= 0x0600`. Putting `#define _WIN32_WINNT 0x0600` at the top of `tool/uriparse.c` (just for debugging) allowed me to trigger a `conflicting types` error now, locally and in CI.

On the base of all that, I would suggest we do two things:

- Use guard `#if defined(__MINGW32__) && (! defined(__MINGW64_VERSION_MAJOR) || __MINGW64_VERSION_MAJOR < 3 || _WIN32_WINNT < 0x0600)` for the future.
  Either with `0x0600` or `_WIN32_WINNT_VISTA`, as you prefer.

- Adjust `socklen_t` to be `size_t` in the prototype.

What do you think?

1480c1: > Use guard `#if defined(__MINGW32__) && (! defined(__MINGW64_VERSION_MAJOR) || __MINGW64_VERSION_MAJOR < 3 || _WIN32_WINNT < 0x0600)` for the future.


This seems fine to me as well as it should cover all cases too

> Adjust `socklen_t` to be `size_t` in the prototype.

👍

hartwork (approved):
@brechtsanders @manisandro any concerns?

hartwork:
PS: @1480c1 very nice commit message :+1:

manisandro:
LGTM, thanks

brechtsanders:
Looks good to me too, thanks for picking this up so fast!

hartwork:
@manisandro thank you!

@brechtsanders thank you! I'm happy about the collaboration, too. Are you good with closing pull request #133 in favor of #131?

I'll wait for @jibsen to review/comment for a few more days since I CC'ed him explicitly and due to his past pull request #53. Great to finally have all those `inet_ntop` pieces come together.

brechtsanders:
@hartwork Sure, #133 can be closed as it's already fixed here.

jibsen:
LGTM, cheers.

hartwork:
Cool, thanks everyone! :tada: :pray: 

Note to self: my now-hidden summary can be found at https://github.com/uriparser/uriparser/pull/131#discussion_r780476071

PR #132, opened 2022-01-07
Fix implicit declaration warning for reallocarray on NetBSD

niclasr:
Fix implicit declaration warning of function reallocarray on NetBSD.


hartwork:
@niclasr thank you!

PR #133, opened 2022-01-07
Fix tool/uriparse.c for MinGW-w64

brechtsanders:
Fix `tool/uriparse.c` for MinGW-w64 where `inet_ntop` is already defined.

### DIFF
--- a/tool/uriparse.c
+++ b/tool/uriparse.c
@@ -43,7 +43,7 @@
 #ifdef _WIN32
 # include <winsock2.h>
 # include <ws2tcpip.h>
-# ifdef __MINGW32__
+# if defined(__MINGW32__) && !defined(InetNtopA)
 WINSOCK_API_LINKAGE const char * WSAAPI inet_ntop(
 		int af, const void *src, char *dst, socklen_t size);
 # endif


hartwork (commented):
Hi @brechtsanders,

@1480c1 and I are already discussing the future of that very source line in pull request #131. Could you join us over there? I'd be curious to learn more about the idea behind `!defined(InetNtopA)`. Thank you! :pray:

tool/uriparse.c
```diff
@@ -43,7 +43,7 @@
 #ifdef _WIN32
 # include <winsock2.h>
 # include <ws2tcpip.h>
-# ifdef __MINGW32__
+# if defined(__MINGW32__) && !defined(InetNtopA)
 WINSOCK_API_LINKAGE const char * WSAAPI inet_ntop(
 		int af, const void *src, char *dst, socklen_t size);
```
1480c1, line 48: You may want to introduce the same define mingw-w64 uses so !defined(InetNtopA) doesn't get triggered again, this is in the case that this ever potentially gets moved to a header file or somewhere else where it could be used multiple times

hartwork:
Closing as replaced by #131

Issue #134, opened 2022-01-11
Failure to Compile: Missing <config.h> CMake Android Target


danoli3:
Not sure why this file is missing... is there any documentation to setting one up otherwise??

Or what in CMake is failing 

Error in build:
```
/Users/~/Library/Android/sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/darwin-x86_64/bin/clang --target=aarch64-none-linux-android21 --sysroot=/Users/~/Library/Android/sdk/ndk/23.1.7779620/toolchains/llvm/prebuilt/darwin-x86_64/sysroot  -I/Users/~/SOURCE/apothecary/apothecary/build/uriparser/src -I/Users/~/SOURCE/apothecary/apothecary/build/uriparser/include -DANDROID -fdata-sections -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -D_FORTIFY_SOURCE=2 -Wformat -Werror=format-security -fexceptions -O3 -DNDEBUG -fPIC -nostdlib -fPIC -MD -MT CMakeFiles/uriparser.dir/src/UriMemory.c.o -MF CMakeFiles/uriparser.dir/src/UriMemory.c.o.d -o CMakeFiles/uriparser.dir/src/UriMemory.c.o -c /Users/one/SOURCE/apothecary/apothecary/build/uriparser/src/UriMemory.c
/Users/~/SOURCE/apothecary/apothecary/build/uriparser/src/UriMemory.c:45:10: fatal error: 'config.h' file not found
#include <config.h>
         ^~~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/uriparser.dir/src/UriMemory.c.o] Error 1
make[1]: *** [CMakeFiles/uriparser.dir/all] Error 2
make: *** [all] Error 2
```

Potential cause:
**config.h** is in the gitignore:
<img width="792" alt="Screen Shot 2022-01-11 at 5 19 29 pm" src="https://user-images.githubusercontent.com/830748/148892798-23b4fc3e-b665-42e2-a592-138b1d921da0.png">

Temp solution remove from UriMemory.h 
```
#include <config.h>
```

Adding PR

hartwork:
Hi @danoli3 ,

> Not sure why this file is missing... is there any documentation to setting one up otherwise??

the file is generated by CMake and put into the build directory, likely `./build/` or `./cmake-build-debug/` in your case.
[...]

PR #135, opened 2022-01-11
Fix missing include config.h

danoli3:
Fixes issue: https://github.com/uriparser/uriparser/issues/134

With missing config.h
Advise if another issues otherwise

This complies successfully


hartwork:
Closing due to https://github.com/uriparser/uriparser/issues/134#issuecomment-1010071495

PR #139, opened 2022-03-03
CMake: Add option to expose (otherwise hidden) symbols

oshogbo:
We would like to propose a new option to CMake, which allows exposing functions to the linker.

Some project like https://github.com/google/sandboxed-api requires this to build a sandbox.

By default, the option will still be disabled, but it would be good to allow the user to change this setting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ option(URIPARSER_BUILD_TESTS "Build test suite (requires GTest >=1.8.0)" ON)
 option(URIPARSER_BUILD_TOOLS "Build tools (e.g. CLI \"uriparse\")" ON)
 option(URIPARSER_BUILD_CHAR "Build code supporting data type 'char'" ON)
 option(URIPARSER_BUILD_WCHAR_T "Build code supporting data type 'wchar_t'" ON)
+option(URIPARSER_VISIBILITY_HIDDEN "Build libraries with hidden symbols unless they are specifically exported" ON)
 option(URIPARSER_ENABLE_INSTALL "Enable installation of uriparser" ON)
 option(URIPARSER_WARNINGS_AS_ERRORS "Treat all compiler warnings as errors" OFF)
 set(URIPARSER_MSVC_RUNTIME "" CACHE STRING "Use of specific runtime library (/MT /MTd /MD /MDd) with MSVC")
@@ -104,9 +105,11 @@ endmacro()
 #
 set(URIPARSER_EXTRA_COMPILE_FLAGS)
 
+if(URIPARSER_VISIBILITY_HIDDEN)
 check_c_compiler_flag("-fvisibility=hidden" URIPARSER_COMPILER_SUPPORTS_VISIBILITY)
 if(URIPARSER_COMPILER_SUPPORTS_VISIBILITY)
     set(URIPARSER_EXTRA_COMPILE_FLAGS "${URIPARSER_EXTRA_COMPILE_FLAGS} -fvisibility=hidden")
+endif()
 endif()
 
 #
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -104,10 +105,12 @@ endmacro()
 #
 set(URIPARSER_EXTRA_COMPILE_FLAGS)
 
+if(HIDE_SYMBOLS)
 check_c_compiler_flag("-fvisibility=hidden" URIPARSER_COMPILER_SUPPORTS_VISIBILITY)
 if(URIPARSER_COMPILER_SUPPORTS_VISIBILITY)
     set(URIPARSER_EXTRA_COMPILE_FLAGS "${URIPARSER_EXTRA_COMPILE_FLAGS} -fvisibility=hidden")
 endif()
+endif(HIDE_SYMBOLS)
```
hartwork, line 113: Please use plain `endif()` for consistency.

Also please increase indent of the new if's body by 4 spaces.

CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ include(GNUInstallDirs)
 # Configuration
 #
 option(BUILD_SHARED_LIBS "Build shared libraries (rather than static ones)" ON)
+option(HIDE_SYMBOLS "Build a libraries with hidden symbols unless they're specifically exported" ON)
```
hartwork, line 62: Please adjust as following:
- Please rename to `URIPARSER_VISIBILITY_HIDDEN`
- Please replace `a libraries` by `libraries`
- Please replace `they're` by `they are`
- Please move the new option right after existing option `URIPARSER_ENABLE_INSTALL`
- Please sync the section in the readme that is auto-listing availbale CMake options; the related shell command is documented at https://github.com/uriparser/uriparser/blame/master/README.md#L51 to make that easier

hartwork (changes requested):
Hi @oshogbo, symbols hidden right now will be called by the API functions, internally. From what I read about sandboxed-api, your application will end up calling the sandbox which will call uriparser for you. That seems to mean that you either will never call non-API functions OR sandboxed-api will magically wrap inner library calls and hide that from you altogether. Is that's what happening? On one hand, I'm happy to add an option like that if it's really needed but on the other I would want to be sure that there is no misunderstanding here and that we'll introduce a zombie option that turns out not needed. If you're 100% that there is inner-library call wrapping happening, please share a link for me to understand the matter and then please check the comments about PR details below to get things in mergable shape. Thank you! :pray:

CMakeLists.txt
oshogbo: Fixed thank you.

oshogbo:
While we compile the libraries with the `visibility=hidden` even the official API is marked as a hidden symbols.
Because of that we can't link to them:
```
[528/556] Linking CXX executable sapi_uriparser.bin
/usr/bin/ld.gold: warning: Cannot export local symbol 'uriEscapeA'
```

With visibility=hidden:
```
$$ objdump -t ./_deps/uriparser-build/liburiparser.a | grep uriParseUriA
00000000000030ae g     F .text  000000000000004d .hidden uriParseUriA
```

Without:
```
00000000000030ae g     F .text  000000000000004d .hidden uriParseUriA
```

hartwork:
@oshogbo PS: I thing I found at least part of the problem. The code at…

https://github.com/uriparser/uriparser/blob/f526f05e828fcd1d03c9ee61d6ce0b8ec002daa6/include/uriparser/UriBase.h#L100-L119

…currently considers visibility to be a thing for shared libraries only. I think that code needs adjustment without breaking things for Windows. Maybe then things already work as expected.

How do we best continue with this? I'm yet not sure how to best fix that code (or what we would actually need here).

hartwork:
@oshogbo for my current draft please see PR #141

oshogbo:
Yes we have to use

oshogbo:
Ups sorry I did some weird stuff.

PR #141, opened 2022-03-16
Fix symbol visibility for non-MSVC static library build (fixes #139)

hartwork:
Fixes #139

CC @oshogbo


oshogbo:
I have just tested and this patch also seems to solve the problem:

`
$ objdump -t ./_deps/uriparser-build/liburiparser.a | grep uriParseUriA
00000000000030ae g     F .text  000000000000004d uriParseUriA
$ ./test/sapi_uriparser_test
[ ... ]
[  PASSED  ] 255 tests.
`

hartwork:
> I have just tested and this patch also seems to solve the problem:

@oshogbo thank you!

PR #149, opened 2022-05-19
Include config.h only when it has not already been defined

gaspardpetit:
This guard makes it possible to build the entire project without a config.h file, instead all the preprocessor definitions can be provided from the command line.

In my case, we build the entire library by including all `.c` files in a single `.c` file with all the needed definitions, thus simplifying the build process:

in `liburiparser.c`
```
#define URI_CONFIG_H 1
#define PACKAGE_VERSION "0.9.6"
#define HAVE_WPRINTF
#include "uriparser/src/UriCommon.c"
#include "uriparser/src/UriCompare.c"
#include "uriparser/src/UriEscape.c"
#include "uriparser/src/UriFile.c"
#include "uriparser/src/UriIp4.c"
#include "uriparser/src/UriIp4Base.c"
#include "uriparser/src/UriMemory.c"
#include "uriparser/src/UriNormalize.c"
#include "uriparser/src/UriNormalizeBase.c"
#include "uriparser/src/UriParse.c"
#include "uriparser/src/UriParseBase.c"
#include "uriparser/src/UriQuery.c"
#include "uriparser/src/UriRecompose.c"
#include "uriparser/src/UriResolve.c"
#include "uriparser/src/UriShorten.c"
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,5 @@
 /.idea/
-/config.h
+/UriConfig.h
 /build/
 /cmake-build-debug/
 /CMakeCache.txt
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,11 +110,11 @@ if(URIPARSER_COMPILER_SUPPORTS_VISIBILITY)
 endif()
 
 #
-# config.h
+# UriConfig.h
 #
 check_symbol_exists(wprintf wchar.h HAVE_WPRINTF)
 check_function_exists(reallocarray HAVE_REALLOCARRAY)  # no luck with CheckSymbolExists
-configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/UriConfig.h.in config.h)
+configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/UriConfig.h.in UriConfig.h)
 
 #
 # C library
@@ -193,7 +193,7 @@ target_include_directories(uriparser
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
     PRIVATE
-        ${CMAKE_CURRENT_BINARY_DIR}  # for config.h
+        ${CMAKE_CURRENT_BINARY_DIR}  # for UriConfig.h
 )
 
 uriparser_install(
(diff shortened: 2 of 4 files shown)


hartwork (commented):
Hi @gaspardpetit I would like to understand your side of things better.

If you're adding a new file `liburiparser.c` with the lines…
```
#define URI_CONFIG_H 1
#define PACKAGE_VERSION "0.9.6"
#define HAVE_WPRINTF
```
today, is anything standing in the way of adding a new file `config.h` with lines…
```
#define PACKAGE_VERSION "0.9.6"
#define HAVE_WPRINTF
```
…and then dropping `URI_CONFIG_H` and the need for patching, altogether?

gaspardpetit:
Hi @hartwork 

To answer your question, no but yes... In a large project, adding a `config.h` file for an external library like this one is risky because multiple third party might have a `config.h` file (it's a pretty common name). The second issue is that the include is using the system include style rather than project include style, i.e.

```
#include <config.h>
```

Instead of

```
#include "config.h"
```

In most projects, there is virtually no difference between the two styles, but in the one where I hit this issue, the include properties are a bit stricter and `<config.h>` won't find the file if it's sitting next to the one including it. Using `"config.h"` instead of `<config.h>` would make it easier for me to have a `config.h` next to my `.c` file, knowing that even if multiple `config.h` files exist in my project, this one would be picked in priority.

I mean, none of this is a big deal, I submitted this pull request with a lot of humility thinking it doesn't improve anything greatly, but it doesn't hurt much either. There will be no hard feeling if you do not see the necessity of merging it ;-)

Alternatively, I could submit another one where I just replace the `<config.h>` for `"config.h`" if you'd prefer that.

Cheers,

Gaspard

hartwork:
Hi @gaspardpetit that's a very nice, detailed and helpful reply, thank you! :+1: 

We can definitely move from `<header>` to `"header"` includes and also move off name `config.h` to `UriConfig.h`.  Would that address all the issues for your environment?

In detail, I envisions this change (with GNU sed rather than macOS):
```console
git grep -l '<config.h>' | xargs -t sed 's,<config.h>,"config.h",g' -i
git grep -l 'config.h' | xargs -t sed 's,config.h,UriConfig.h,g' -i
```
If you like that approach, we can do that in this pull request, or I can make that a new pull request… — what do you think?

Best, Sebastian

gaspardpetit:
This would definitely work for my use cases, thank you for your consideration!

hartwork:
@gaspardpetit excellent, thanks! :+1:

PR #150, opened 2022-06-03
Fix IPv6 and IPvFuture literal parsing

### LINKED ISSUE #146: Multiple issues with IPv6 and IPvFuture literal parsing
I found several issues with IP literal parsing in `uriparser` when fuzzing [`fluent-uri`](https://github.com/yescallop/fluent-uri-rs) against this library. Here's a list of them:

- According to RFC 3986 (RFC 3513), the use of `::` in an IPv6 address indicates one or more groups of 16 bits of zeros. However, parsing `//[0:0:0:0:0:0:0::1]` succeeds.
- A leading or trailing `:` that is not part of a `::` is not permitted in an IPv6 address, but parsing `//[:1::1:]` succeeds.
- The leading `v` of an IPvFuture is case-insensitive (see RFC 3986), but parsing `//[VF.addr]` fails.
- Parsing `//[::1.1.1.11]` and `//[::1.1.1.111]` succeeds somehow, but parsing `//[::1.1.1.100]` fails.
- Parsing `//[1:1:1:1:1:1::1.1.1.1]` succeeds.

Here's my fuzz target against `uriparser` which I hope could help: [against_uriparser.rs](https://github.com/yescallop/fluent-uri-rs/blob/main/fuzz/fuzz_targets/against_uriparser.rs).

yescallop:
Closes #146. This is ready for review now. @hartwork

I've added several test cases for the issue and can confirm that all the five bugs listed in #146 are fixed.

### DIFF
--- a/src/UriParse.c
+++ b/src/UriParse.c
@@ -782,7 +782,7 @@ static const URI_CHAR * URI_FUNC(ParseIPv6address2)(
 								return NULL; /* ":::+ "*/
 							}
 						} else if (quadsDone == 0 || first[1] == _UT(']')) {
-							/* Leading or trailing ":" */
+							/* Single leading or trailing ":" */
 							URI_FUNC(StopSyntax)(state, first, memory);
 							return NULL;
 						}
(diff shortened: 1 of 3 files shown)


yescallop:
I left the fuzzer running and it eventually found another failing case. Just fixed that!

hartwork (changes requested):
@yescallop excellent work, thanks for the pull request and for diving into the code this far :+1: A few comments on details only, below. Thanks in advance!

hartwork (approved):
@yescallop very nice, thanks for the adjustments! :pray:

PR #170, opened 2023-06-15
CMake: Introduce option `URIPARSER_SHARED_LIBS=(ON|OFF)` (fixes #169)

### LINKED ISSUE #169: CMake: Please introduce an option URIPARSER_SHARED_LIBS to allow fine-grained control over BUILD_SHARED_LIBS
Hi,

I was recently looking at incorporating this project iinto another via using [`FetchContent`](https://cmake.org/cmake/help/latest/module/FetchContent.html).
However I ran into some problems with [`BUILD_SHARED_LIBS`](https://github.com/uriparser/uriparser/blob/56213158a9a705628e7b10cc840cd161cdea5b48/CMakeLists.txt#L61).

As this is a cache entry its global, and had the unfortunately effect of building everything else in the project as shared.
The build is currently _mostly_ static libraries by default, with the third party libraries such as uriparser being the exception.

So, I was wondering what peoples thoughts would be on some changes to the cmake system to allow the library to be easily incorporated into a larger project?

I wouldn't be particularly comfortable with CMake, but skimming the file I think `BUILD_SHARED_LIBS` is the only thing which stands out to me. But I'm not entirely sure how to best manage that.

- Only add it to the cache if `uriparser` is the top level project?
- Juggle some `URIPARSER_BUILD_SHARED_LIBS` variable instead?

Regards,
Alan

hartwork:
.. that mimics flag `BUILD_SHARED_LIBS` and falls back to `BUILD_SHARED_LIBS` if available, else defaults to `ON`.

Related:
https://cmake.org/cmake/help/latest/variable/BUILD_SHARED_LIBS.html

Fixes #169

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,12 @@ include(GNUInstallDirs)
 #
 # Configuration
 #
-option(BUILD_SHARED_LIBS "Build shared libraries (rather than static ones)" ON)
+if(DEFINED BUILD_SHARED_LIBS)
+    set(_URIPARSER_SHARED_LIBS_DEFAULT ${BUILD_SHARED_LIBS})
+else()
+    set(_URIPARSER_SHARED_LIBS_DEFAULT ON)
+endif()
+option(URIPARSER_SHARED_LIBS "Build shared libraries (rather than static ones)" ${_URIPARSER_SHARED_LIBS_DEFAULT})
 option(URIPARSER_BUILD_DOCS "Build API documentation (requires Doxygen, Graphviz, and (optional) Qt's qhelpgenerator)" ON)
 option(URIPARSER_BUILD_TESTS "Build test suite (requires GTest >=1.8.0)" ON)
 option(URIPARSER_BUILD_TOOLS "Build tools (e.g. CLI \"uriparse\")" ON)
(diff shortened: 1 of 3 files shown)


alanbarr:
Performed some quick tests - looks good to me.

# Setup

Testing with FetchContent arguments:
```  
GIT_REPOSITORY https://github.com/uriparser/uriparser.git
GIT_TAG f79d49f4e8b94b5691e0611054541923db7c3b7d
```

# Only Changing URIPARSER_SHARED_LIBS

For a static build with only the following set prior to FetchContent:
```
option(URIPARSER_SHARED_LIBS "" OFF)
```
We get:
```
-- Configuration
--   Build type ............. debug
--   Shared libraries ....... OFF
```
And
```
liburiparser.a
```

For a shared build with only the following set prior to FetchContent (and a cleaned up Cache)

```
option(URIPARSER_SHARED_LIBS "" ON)
```
We get:

```
-- Configuration
--   Build type ............. debug
--   Shared libraries ....... ON
```
and:
```
liburiparser.so.1.0.30
```

# BUILD_SHARED_LIBS / BOTH

For a static build with only the following set prior to FetchContent (and a cleaned up Cache):
```
option(BUILD_SHARED_LIBS "" OFF)
```
We get:
```
-- Configuration
--   Build type ............. debug
--   Shared libraries ....... OFF
```
and
```
liburiparser.a
```


Editing the Cache to have:
```
BUILD_SHARED_LIBS:BOOL=OFF
URIPARSER_SHARED_LIBS:BOOL=ON
```

We get:
```
-- Configuration
--   Build type ............. debug
--   Shared libraries ....... ON
--   Compiler flags
```
and
```
liburiparser.so.1.0.30
```

hartwork:
@alanbarr excellent, thanks for the testing! I'll go ahead and merge in a second. Please open a new issue for shortcomings about it coming to light after merging if any. Thanks! :pray:

PR #197, opened 2024-09-26
Add alias uriparser::uriparser

Vollstrecker:
uriparser is exported as uriparser::uriparser so I added this alias here also. Otherwise devs would have to differentiate if they found the lib with find_package or are chainbuilding with i.e. fetchcontent

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -167,6 +167,8 @@ add_library(uriparser
     ${LIBRARY_CODE_FILES}
 )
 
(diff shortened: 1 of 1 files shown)


hartwork:
@Vollstrecker I'm not sure I get the need or the idea, similar to my reply at https://github.com/libexpat/libexpat/pull/903#pullrequestreview-2332341728 , but let's only discuss there or here, so that we don't duplicate the thread more. Thank you!

Vollstrecker:
Sure, I've seen the other one first.

Issue #32, opened 2018-09-27
Start tracking (or at least reporting on) code coverage(, integrate with CI)


hartwork:
Please get in touch about details before starting work on this matter. Thanks!

Issue #16, opened 2018-07-11
Incomplete test coverage for UriFile.c


schwehr:
I redid the tests for `UriFile.c` for my own use in GoogleTest.  In doing so, I noticed that `test.cpp` doesn't fully cover the compilation unit.  You are welcome to take any of my test code and adapt for uriparser.  If you aren't interested in the coverage, feel free to close the issue.  

The checks with CheckUriUriStringToWindowsFilenameA are bit chaotic and could use some improvement.  The code is only tested with ASCII input as I don't build with wchar_t support.

I do one test file for each compilation unit rather than a catch all test.cpp.

Code is copyright Google and donated to uriparser under the uriparser license. 

Here is what I have:

```c++
// Test UriFile.c
//
// A = ASCII
//
// Many of these test examples show very strange behavior.

#include <cstddef>
#include <cstring>
#include <vector>

#include "gtest.h"
#include "uriparser/Uri.h"
#include "/uriparser/UriBase.h"

namespace {

TEST(UriUnixFilenameToUriStringA, Nullptr) {
  const char uri[] = "/bin/bash";
  std::vector<char> buf(10);

  ASSERT_EQ(URI_ERROR_NULL, uriUnixFilenameToUriStringA(nullptr, &buf[0]));
  ASSERT_EQ(URI_ERROR_NULL, uriUnixFilenameToUriStringA(uri, nullptr));
  ASSERT_EQ(URI_ERROR_NULL, uriUnixFilenameToUriStringA(nullptr, nullptr));
}

size_t UnixFilenameToUriStringSize(const char *filename) {
  // "file://" plus each character could be expanded to a 3 character percent
  // representation and finishing up with a NUL termination sentinal.
  return 7 + 3 * strlen(filename) + 1;
}

void CheckUriUnixFilenameToUriStringA(const char *filename, const char *uri) {
  std::vector<char> buf(UnixFilenameToUriStringSize(filename));
  ASSERT_EQ(URI_SUCCESS, uriUnixFilenameToUriStringA(filename, &buf[0]));
  EXPECT_STREQ(uri, &buf[0]) << "For: \"" << filename << "\"";
}

TEST(UriUnixFilenameToUriStringA, WhiteSpace) {
  CheckUriUnixFilenameToUriStringA("", "");
  CheckUriUnixFilenameToUriStringA(" ", "%20");
  CheckUriUnixFilenameToUriStringA("a b", "a%20b");
[...]

PR #209, opened 2025-01-02
Integrate fuzzers from `google/oss-fuzz` repository + improve those fuzzers on top

tyler92:
Fuzz targets improved and moved from [google/oss-fuzz](https://github.com/google/oss-fuzz/tree/master/projects/uriparser) repository.

The primary advantage of this change is gaining full ownership of the fuzz targets. This allows for direct modifications to existing fuzz targets and the addition of new ones in the upstream repository, without needing to involve the maintainers of `google/oss-fuzz`. Given that `google/oss-fuzz` hosts hundreds of projects, this change significantly reduces the burden on its maintainers.

The main enhancement introduced in this PR is the extension of fuzzing tests to include Unicode functions. Specifically, each fuzz target now has a corresponding `w` version. As a result, fuzzing tests achieve significantly better coverage (89% compared to 45% previously). This improvement may help address issues such as https://github.com/uriparser/uriparser/issues/32 and probably https://github.com/uriparser/uriparser/issues/16. I tried to avoid other big changes to make it easy to compare `cpp` files between two repositories.

If this PR is approved and merged, I will submit a follow-up PR to the `google/oss-fuzz` repository to "activate" it.


hartwork:
@tyler92 thanks for your pull request. It will take a few days for me to find time for proper review.

hartwork (commented):
@tyler92 I had a chance to look through this a bit more. I understand the general idea and how it increases coverage. Before we go into details, my key concern and question at the moment is if we're even allowed to re-license Apache-licensed code under LGPL — I'm not sure that works. Is sticking to the Apache license an option to you? Would work for me.

If yes, let's:
- please make this at least two commits, the first adding unmodified fuzzer code files from https://github.com/google/oss-fuzz/tree/master/projects/uriparser and the second making adjustments to that code.
- please put the code at `./fuzz` rather than `./test/fuzzing`. It's not unit tests, the code will be easier to find, shorter paths to write also.

What do you think?

hartwork:
PS: Block…

> You must retain, in the Source form of any Derivative Works that You distribute, all copyright, patent, trademark, and attribution notices from the Source form of the Work, excluding those notices that do not pertain to any part of the Derivative Works; and

…at https://www.apache.org/licenses/LICENSE-2.0 would seem to dictate that e.g. we keep the "Copyright 2020 Google LLC" intact (and add your name to it).

tyler92:
Hi @hartwork 

> Is sticking to the Apache license an option to you? Would work for me.

Yes, sure, I will revert to Apache. Sorry, It's my bad. I wasn't focused on the license aspect.

> - please make this at least two commits, the first adding unmodified fuzzer code files from https://github.com/google/oss-fuzz/tree/master/projects/uriparser and the second making adjustments to that code.
> - please put the code at ./fuzz rather than ./test/fuzzing. It's not unit tests, the code will be easier to find, shorter paths to write also.

Sure, I will probably make 3 commits: adding unmodified code, formatting, and new changes. I tried to make formatting similar to other code in this repository, including files naming - CamelCase instead of snake_case and `.cpp` instead of `.cc`.

hartwork:
@tyler92 cool, thank you! :+1:

tyler92:
I pushed four commits:

- Added unmodified code.
- Applied formatting.
- Renamed fuzz target files (to separate renaming from modifying files in the same commit).
- Introduced new changes, including adding my name to the copyright notice. Please note that `FuzzingUtils.h` is a new file, so the license includes only my name.

We can squash these commits after the review if you prefer. For now, I've kept them separate to make the review process easier.

hartwork:
> I pushed four commits:

@tyler92 thanks!

> * Added unmodified code.

I haven't mentioned this yet but I think we do not need the `.sh` build script around here at all. Can we drop that file from all commits? Some details seem off and I don't see any upstream value in the file.

> * Applied formatting.

The commit does a lot more than change formatting. I think my favorite would be (a) to not change formatting for the smallest possible diff to the original, e.g. to ease review. Another option would be (b) for me to configure clang-format CI for the whole code base and then we re-use that setting. Or we (c) make this commit to be pure clang-format adjustments and nothing more (with a clang-format preset of your choice). What do out think?

> * Renamed fuzz target files (to separate renaming from modifying files in the same commit).

:+1: 

> * Introduced new changes, including adding my name to the copyright notice. Please note that `FuzzingUtils.h` is a new file, so the license includes only my name.

:+1: (in general)

> We can squash these commits after the review if you prefer. For now, I've kept them separate to make the review process easier.

Separate commits is great (as long as we don't have chains of "fix x", "fix the fix for x" and so on :smile: )

tyler92:
> I haven't mentioned this yet but I think we do not need the .sh build script around here at all. Can we drop that file from all commits? Some details seem off and I don't see any upstream value in the file.

Done

> The commit does a lot more than change formatting. I think my favorite would be (a) to not change formatting for the smallest possible diff to the original, e.g. to ease review.

Hmm, it is pure clang-format adjustments with two manual changes: (a) three empty lines between functions, (b) adding `{}` around single-line if-s. Do you prefer to move these two manual changes to another commit? Probably the diff looks huge because two spaces were replaced with tabs. Just for reference: I used the following clang-format config: [clang-format.txt](https://github.com/user-attachments/files/18307705/clang-format.txt)

hartwork:
> > I haven't mentioned this yet but I think we do not need the .sh build script around here at all. Can we drop that file from all commits? Some details seem off and I don't see any upstream value in the file.
> 
> Done

@tyler92 thanks!


> > The commit does a lot more than change formatting. I think my favorite would be (a) to not change formatting for the smallest possible diff to the original, e.g. to ease review.
> 
> Hmm, it is pure clang-format adjustments with two manual changes: (a) three empty lines between functions, (b) adding `{}` around single-line if-s. Do you prefer to move these two manual changes to another commit? Probably the diff looks huge because two spaces were replaced with tabs. Just for reference: I used the following clang-format config: [clang-format.txt](https://github.com/user-attachments/files/18307705/clang-format.txt)

My bad, I was too quick with that assessment. It's probably already to have the two if body things in there.

Up next: Actual review :smile:

hartwork:
@tyler92 regarding the commit messages:

- "Code formatiing for fuzz targets" has typo "formatiing"

- "Improve fuzzing targets by covering Unicode functions" makes me want to mention that the whole use of term Unicode in uriparser was mistaken back then, it's wide character support, there is no unicode support in here.  Just so you know.

hartwork:
@tyler92 any idea how to figure out the public corpus download URL? For Expat it's https://storage.googleapis.com/expat-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/expat_xml_parse_fuzzer_UTF-8/public.zip (and a few like that). I've been trying things like https://storage.googleapis.com/uriparser-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/uriparser_uri_free/public.zip but so far without success :thinking:

tyler92:
@hartwork Corrected commit messages

The correct URL for the public corpus is https://storage.googleapis.com/uriparser-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/uriparser_uri_free_fuzzer/public.zip - the part before "public.zip" is the fuzzer name. Existing names:

```
uri_dissect_query_malloc_fuzzer
uri_free_fuzzer
uri_parse_fuzzer
```
New ones:

```
uri_dissect_query_mallocw_fuzzer
uri_freew_fuzzer
uri_parsew_fuzzer
```

CMakeLists.txt
```diff
@@ -334,6 +335,29 @@ if(URIPARSER_BUILD_TESTS)
     add_custom_target(check COMMAND ${CMAKE_CTEST_COMMAND})
 endif()
 
+#
+# Fuzzers
+#
+if(URIPARSER_BUILD_FUZZERS AND DEFINED ENV{LIB_FUZZING_ENGINE})
```
hartwork, line 341: I would rather not want to error out without `LIB_FUZZING_ENGINE` but to link with `-fsanitize=fuzzer` instead for users outside of OSS-Fuzz. For Expat, there is a dedicated flag `EXPAT_OSSFUZZ_BUILD` that allows users tselect from those two flavors of fuzzing:

- https://github.com/libexpat/libexpat/blob/c1c1677657944cf15faf59374c5b67f0541246f9/expat/CMakeLists.txt#L723-L729
- https://github.com/libexpat/libexpat/blob/c1c1677657944cf15faf59374c5b67f0541246f9/expat/CMakeLists.txt#L735-L737
- https://github.com/libexpat/libexpat/blob/c1c1677657944cf15faf59374c5b67f0541246f9/expat/CMakeLists.txt#L746-L755

How do you feel about a similar approach with a second flag `URIPARSER_OSSFUZZ_BUILD` in general?

CMakeLists.txt
```diff
@@ -334,6 +335,29 @@ if(URIPARSER_BUILD_TESTS)
...
+        if(URIPARSER_BUILD_WCHAR_T)
+            add_executable(${name}w_fuzzer ${source})
+            target_link_libraries(${name}w_fuzzer PRIVATE uriparser $ENV{LIB_FUZZING_ENGINE})
+            target_compile_definitions(${name}w_fuzzer PRIVATE URI_NO_ANSI)
+        endif()
+    endfunction()
+
+    add_ansi_and_unicode_fuzzer(uri_free fuzz/FreeFuzzer.cpp)
+    add_ansi_and_unicode_fuzzer(uri_parse fuzz/ParseFuzzer.cpp)
+    add_ansi_and_unicode_fuzzer(uri_dissect_query_malloc fuzz/DissectQueryMallocFuzzer.cpp)
```
hartwork, line 358: Alphabetical order please :pray:

CMakeLists.txt
```diff
@@ -495,6 +519,7 @@ message(STATUS "    Code for char * ...... ${URIPARSER_BUILD_CHAR}")
 message(STATUS "    Code for wchar_t * ... ${URIPARSER_BUILD_WCHAR_T}")
 message(STATUS "    Tools ................ ${URIPARSER_BUILD_TOOLS}")
 message(STATUS "    Test suite ........... ${URIPARSER_BUILD_TESTS}")
+message(STATUS "    Fuzzing tests ........ ${URIPARSER_BUILD_FUZZERS}")
```
hartwork, line 522:
```suggestion
message(STATUS "    Fuzzers .............. ${URIPARSER_BUILD_FUZZERS}")
```

(Because these are not tests to me.)

README.md
```diff
@@ -67,6 +67,9 @@ URIPARSER_BUILD_DOCS:BOOL=ON
 // Build test suite (requires GTest >=1.8.0)
 URIPARSER_BUILD_TESTS:BOOL=ON
 
+// Build fuzzers (requires Clang and LIB_FUZZING_ENGINE env)
+URIPARSER_BUILD_FUZZERS:BOOL=ON
```
hartwork, line 71:
```suggestion
URIPARSER_BUILD_FUZZERS:BOOL=OFF
```

As produced by the command above

fuzz/DissectQueryMallocFuzzer.cpp
```diff
@@ -0,0 +1,60 @@
...
+	if (!chars_required) {
+		URI_FUNC(FreeQueryList)(query_list);
+		return 0;
+	}
+
+	std::vector<URI_CHAR> buf(chars_required, 0);
+	int written = -1;
+
+	// Reverse the process of uriDissectQueryMallocA.
+	result = URI_FUNC(ComposeQuery)(buf.data(), query_list, chars_required, &written);
```
hartwork, line 56: These should be using `chars_required + 1`  probably, because [`uriComposeQueryCharsRequiredA` excludes the terminator](https://uriparser.github.io/doc/api/latest/Uri_8h.html#a45b2899b4cdeeb8d229543ea3edfd872) while [`uriComposeQueryA` does include it](https://uriparser.github.io/doc/api/latest/Uri_8h.html#a9b8b24dc611176e3253f8b7ae6d6c737). What do you think?

hartwork (changes requested):
@tyler92 flushing review from first round… Haven't seen all code yet, more coming (though not in the next hour)…

hartwork:
@tyler92 PS: Thanks for the corpus URLs! :pray:

CMakeLists.txt
tyler92: I'm OK with this approach but I want to ask one question before. This approach requires a separate target for the library that will be linked with the fuzzer:

https://github.com/libexpat/libexpat/blob/c1c1677657944cf15faf59374c5b67f0541246f9/expat/CMakeLists.txt#L734

that duplicates the "regular" target with a couple of changes like `target_compile_options(fuzzpat PRIVATE -fsanitize=fuzzer-no-link)`. It means that `uriparser` library will be compiled twice if `URIPARSER_BUILD_FUZZERS=ON`. Are you OK with this?

CMakeLists.txt
tyler92: Fixed

CMakeLists.txt
tyler92: Fixed

README.md
tyler92: Fixed

fuzz/DissectQueryMallocFuzzer.cpp
tyler92: Yes, I agree with you. Fixed

CMakeLists.txt
hartwork: @tyler92 thanks for asking! If our options are…

- a) add `-fsanitize=fuzzer-no-link` to mainline uriparser (and putting it into non-production state) or
- b) have two copies

…I'm for two copies. Are there alternatives? We can call the new target `fuzzparser` if you like :smile:

fuzz/FuzzingUtils.h
```diff
@@ -0,0 +1,50 @@
...
+#ifndef URI_FUZZING_UTILS_H
+#define URI_FUZZING_UTILS_H 1
+
+#include "uriparser/Uri.h"
+#include <fuzzer/FuzzedDataProvider.h>
+#include <string>
+
+
+
+using String = std::basic_string<URI_CHAR>;
```
hartwork, line 24: If this was named `UriString` the risk of mistaking `String` below for `std::string` would reduce.

fuzz/FuzzingUtils.h
```diff
@@ -0,0 +1,50 @@
...
+
+
+
+inline String consumeBytesAsString(FuzzedDataProvider & fdp, size_t len) {
+	String str(len, 0);
+	size_t bytes = fdp.ConsumeData(str.data(), str.length() * sizeof(URI_CHAR));
```
hartwork, line 30:
```suggestion
	size_t bytes = fdp.ConsumeData(str.data(), len * sizeof(URI_CHAR));
```
Would seem simpler if identical. Just an idea.

fuzz/FuzzingUtils.h
```diff
@@ -0,0 +1,50 @@
...
+	str.resize(bytes / sizeof(URI_CHAR));
```
hartwork, line 31: I have a vague idea of the mission here but it would be cool to be sure. Could you add a line of comment about the case and motivation here (and to the sibling case a bit further down? :pray:

fuzz/FuzzingUtils.h
```diff
@@ -0,0 +1,50 @@
...
+	return str;
+}
+
+
+
+inline String consumeRandomLengthString(FuzzedDataProvider & fdp) {
+	String str(fdp.ConsumeIntegralInRange(0ul, fdp.remaining_bytes()), 0);
```
hartwork, line 38: Can we maybe do something like…

```suggestion
        const size_t len = fdp.ConsumeIntegralInRange(0ul, fdp.remaining_bytes());
	String str(len, 0);
```

…to (arguably?) improve readability of that part?

fuzz/ParseFuzzer.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+private:
+	URI_TYPE(Uri) uri_;
+};
+
+
+
+void Escapes(const String & uri) {
+	const URI_CHAR * first = uri.c_str();
+	// A new line char takes 6 char to encode.
```
hartwork, line 47: This is about "%0D%0A" with `normalizeBreaks` enabled, right? Would be cool to enrich this comment a bit to address the natural questions coming up. :pray:

hartwork (changes requested):
@tyler92 found a few smaller (or even tiny) things below:

CMakeLists.txt
tyler92: OK, got it, agree with you. Thank you for the hint :+1:  I will make all necessary changes tomorrow

hartwork: @tyler92 PS: I should note that de-facto code…

```cmake
            if(NOT EXPAT_OSSFUZZ_BUILD)
                target_compile_options(${target_name} PRIVATE -fsanitize=fuzzer-no-link)
                set_target_properties(${target_name} PROPERTIES LINK_FLAGS -fsanitize=fuzzer)
            endif()
```

…as a subset of https://github.com/libexpat/libexpat/blob/c1c1677657944cf15faf59374c5b67f0541246f9/expat/CMakeLists.txt#L746-L755 started puzzling me as self-conflicting and unintended. I might be missing something, I have a feeling that `fuzzer-no-link` should be dropped here because it's the user facing fuzzer binaries this is for. Does it seem right as is to you?

hartwork:
> The correct URL for the public corpus is https://storage.googleapis.com/uriparser-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/uriparser_uri_free_fuzzer/public.zip - the part before "public.zip" is the fuzzer name. Existing names:
> 
> ```
> uri_dissect_query_malloc_fuzzer
> uri_free_fuzzer
> uri_parse_fuzzer
> ```
> 
> New ones:
> 
> ```
> uri_dissect_query_mallocw_fuzzer
> uri_freew_fuzzer
> uri_parsew_fuzzer
> ```

@tyler92 I have used the corpus URLs now to add CI on top of your current work here, 99% matching [what Expat did on the same topic](https://github.com/libexpat/libexpat/blob/master/.github/workflows/fuzzing.yml): download corpora and run the fuzzers against those. My WIP on that is in commit 5f8605be47af9cf7b9641b1743fb55ea1ee7eb52 . I can introduce something like as a second pull request later or you cherry-pick my commit on top. If you think that addition of that commit could endanger or complicate the remuneration of this pull request, the second pull request approach would solve that problem and I can still query CI for approval. Just thinking aloud. What do you think?

CMakeLists.txt
tyler92: I've added `URIPARSER_OSSFUZZ_BUILD` with the same logic as for Expat.

> I have a feeling that fuzzer-no-link should be dropped here because it's the user facing fuzzer binaries this is for. Does it seem right as is to you?

I think it's OK: we compile `.c`/`.cpp` files with `-fsanitize=fuzzer-no-link` (even the main fuzz `cpp` file that contains `LLVMFuzzerTestOneInput`) and link with `-fsanitize=fuzzer`

fuzz/FuzzingUtils.h
tyler92: Renamed

fuzz/FuzzingUtils.h
tyler92: I've refactored FuzzingUtils.h a little bit and applied your suggestion

fuzz/FuzzingUtils.h
tyler92: I've added a comment :ok_hand:

fuzz/FuzzingUtils.h
tyler92: I've refactored FuzzingUtils.h a little bit (to make the difference between `bytes` and `len` more obvious) and applied your suggestion

fuzz/ParseFuzzer.cpp
tyler92: I think yes, the comment is extended

tyler92:
> > The correct URL for the public corpus is https://storage.googleapis.com/uriparser-backup.clusterfuzz-external.appspot.com/corpus/libFuzzer/uriparser_uri_free_fuzzer/public.zip - the part before "public.zip" is the fuzzer name. Existing names:
> > ```
> > uri_dissect_query_malloc_fuzzer
> > uri_free_fuzzer
> > uri_parse_fuzzer
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > New ones:
> > ```
> > uri_dissect_query_mallocw_fuzzer
> > uri_freew_fuzzer
> > uri_parsew_fuzzer
> > ```
> 
> @tyler92 I have used the corpus URLs now to add CI on top of your current work here, 99% matching [what Expat did on the same topic](https://github.com/libexpat/libexpat/blob/master/.github/workflows/fuzzing.yml): download corpora and run the fuzzers against those. My WIP on that is in commit [5f8605b](https://github.com/uriparser/uriparser/commit/5f8605be47af9cf7b9641b1743fb55ea1ee7eb52) . I can introduce something like as a second pull request later or you cherry-pick my commit on top _in here_. If you think that addition of that commit could endanger or complicate the remuneration of this pull request, the second pull request approach would solve that problem and I can still query CI for approval. Just thinking aloud. What do you think?

It's up to you, I can cherry-pick it but I see two things here:

1. It's not possible to uncomment `w` fuzzers there right now because they don't exist yet. They will exist after we merge the current PR
2. There are TODOs and it's probably better if you resolve them first, I can assist you if you need my help

Btw, I noticed that there are already 6 fuzzers here https://introspector.oss-fuzz.com/project-profile?project=uriparser although this PR is not merged yet. Did you add them somehow manually (this part of OSS-Fuzz is a block box to me).

fuzz/FuzzingUtils.h
hartwork: @tyler92 thanks! Could you add the same comment to the second `.resize` right here, also?
I'm also wondering why it's okay to consume fewer bytes and that maybe the function should be named something like `tryConsumeBytesAsString` with emphasis on trying.

fuzz/ParseFuzzer.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+private:
+	URI_TYPE(Uri) uri_;
+};
+
+
+
+void Escapes(const UriString & uri) {
+	const URI_CHAR * first = uri.c_str();
+	// Up to 6 bytes per character with normalizeBreaks enabled (\\n -> %0D%0A)
```
hartwork, line 47:
```suggestion
	// Up to 6 bytes per character with normalizeBreaks enabled (\n -> %0D%0A)
```
The double backslash seems mistaken (or I didn't get the idea)

CMakeLists.txt
hartwork: @tyler92 I don't get it — why would we do that? I would expect them to ideally match.

CMakeLists.txt
```diff
@@ -66,6 +66,8 @@ endif()
 option(URIPARSER_SHARED_LIBS "Build shared libraries (rather than static ones)" ${_URIPARSER_SHARED_LIBS_DEFAULT})
 option(URIPARSER_BUILD_DOCS "Build API documentation (requires Doxygen, Graphviz, and (optional) Qt's qhelpgenerator)" ON)
 option(URIPARSER_BUILD_TESTS "Build test suite (requires GTest >=1.8.0)" ON)
+option(URIPARSER_BUILD_FUZZERS "Build fuzzers (requires Clang)" OFF)
+option(URIPARSER_OSSFUZZ_BUILD "Build fuzzers via ossfuzz" OFF)
```
hartwork, line 70:
```suggestion
option(URIPARSER_OSSFUZZ_BUILD "Build fuzzers via OSS-Fuzz" OFF)
```

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+        message(FATAL_ERROR
+            "Building fuzz targets without Clang (but ${CMAKE_C_COMPILER_ID}) is not "
+            "supported. Please set -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++.")
+    endif()
+
+    if(URIPARSER_OSSFUZZ_BUILD AND NOT DEFINED ENV{LIB_FUZZING_ENGINE})
+        message(FATAL_ERROR
+            "OSS-Fuzz builds require the environment variable "
+            "LIB_FUZZING_ENGINE to be set. If you are seeing this "
+            "warning, it points to a deeper problem in the ossfuzz "
```
hartwork, line 353:
```suggestion
            "warning, it points to a deeper problem in the OSS-Fuzz "
```
(I should fix this in Expat, too.)

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+        endif()
+    endfunction()
+
+    add_ansi_and_unicode_fuzzer(uri_dissect_query_malloc fuzz/DissectQueryMallocFuzzer.cpp)
+    add_ansi_and_unicode_fuzzer(uri_free fuzz/FreeFuzzer.cpp)
+    add_ansi_and_unicode_fuzzer(uri_parse fuzz/ParseFuzzer.cpp)
+else()
+    if(URIPARSER_OSSFUZZ_BUILD)
+        message(SEND_ERROR
+                "Attempting to perform an ossfuzz build without turning on the fuzzer build. "
```
hartwork, line 399:
```suggestion
                "Attempting to perform an OSS-Fuzz build without turning on the fuzzer build. "
```

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+                "This is likely in error - consider adding "
+                "-DURIPARSER_BUILD_FUZZERS=ON to your cmake execution.")
```
hartwork, line 401:
```suggestion
                "-DURIPARSER_BUILD_FUZZERS=ON to your CMake execution.")
```

README.md
```diff
@@ -67,6 +67,12 @@ URIPARSER_BUILD_DOCS:BOOL=ON
 // Build test suite (requires GTest >=1.8.0)
 URIPARSER_BUILD_TESTS:BOOL=ON
 
+// Build fuzzers (requires Clang)
+URIPARSER_BUILD_FUZZERS:BOOL=OFF
+
+// Build fuzzers via ossfuzz
```
hartwork, line 73:
```suggestion
// Build fuzzers via OSS-Fuzz
```

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+        target_link_libraries(${target_name} PRIVATE fuzzparser)
+        target_compile_definitions(${target_name} PRIVATE ${compile_definitions})
+        if(URIPARSER_OSSFUZZ_BUILD)
+            set_target_properties(${target_name} PROPERTIES LINK_FLAGS $ENV{LIB_FUZZING_ENGINE})
+            set_target_properties(${target_name} PROPERTIES LINKER_LANGUAGE "CXX")
+        else()
+            target_compile_options(${target_name} PRIVATE -fsanitize=fuzzer-no-link)
+            set_target_properties(${target_name} PROPERTIES LINK_FLAGS -fsanitize=fuzzer)
+        endif()
+    endfunction()
```
hartwork, line 381:
```suggestion
        endif()
        set_property(TARGET ${target_name} PROPERTY RUNTIME_OUTPUT_DIRECTORY fuzz)        
    endfunction()
```
(It's a detail, but would be nice. Matches Expat.)

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+
+    function (add_ansi_and_unicode_fuzzer name source)
```
hartwork, line 383:
```suggestion
    function(add_ansi_and_unicode_fuzzer name source)
```

CMakeLists.txt
```diff
@@ -334,6 +336,72 @@ if(URIPARSER_BUILD_TESTS)
...
+            ${CMAKE_CURRENT_SOURCE_DIR}/include
+        PRIVATE
+            ${CMAKE_CURRENT_BINARY_DIR}  # for UriConfig.h
+    )
+
+    if(NOT URIPARSER_OSSFUZZ_BUILD)
+        target_compile_options(fuzzparser PRIVATE -fsanitize=fuzzer-no-link)
+    endif()
+
+    function (add_fuzzer target_name source compile_definitions)
```
hartwork, line 370:
```suggestion
    function(add_fuzzer target_name source compile_definitions)
```

hartwork (changes requested):
@tyler92 looks pretty good to me already, a few details that I found more:

CMakeLists.txt
tyler92: Corrected

CMakeLists.txt
tyler92: Corrected

CMakeLists.txt
tyler92: Added

README.md
tyler92: Corrected

CMakeLists.txt
tyler92: Corrected

CMakeLists.txt
tyler92: Corrected

CMakeLists.txt
tyler92: Corrected

CMakeLists.txt
tyler92: Corrected

fuzz/ParseFuzzer.cpp
tyler92: Fixed

hartwork:
> It's up to you, I can cherry-pick it but I see two things here:
> 
> 1. It's not possible to uncomment `w` fuzzers there right now because they don't exist yet. They will exist after we merge the current PR

@tyler92 understood, yes. That's the only TODO that will have to remain for now. I'll take care of it once the other three corpora appear online.


> 2. There are TODOs and it's probably better if you resolve them first, I can assist you if you need my help

That's fair, I resolve all else. Please cherry-pick e62e6ea1a7c8ea8ae09be6f69ad4c60d7c2957d1. They only changed needed more on the side of your commits would be this: 4998d724155e504db681a0d9fb6c7a896babd496. if you squash that into the right commit, that would be perfect.


> Btw, I noticed that there are already 6 fuzzers here https://introspector.oss-fuzz.com/project-profile?project=uriparser although this PR is not merged yet. Did you add them somehow manually (this part of OSS-Fuzz is a block box to me).

I did not touch anything there, not sure.

fuzz/FuzzingUtils.h
tyler92: Now `consumeRandomLengthString` calls `tryConsumeBytesAsString` (I renamed it) so only one comment is needed. 

> I'm also wondering why it's okay to consume fewer bytes

It's a common practice for fuzzing. Actually, these three functions are wrappers for `FuzzedDataProvider`'s methods with the same names. It supports only `std::string` but not `std::wstring`. And `FuzzedDataProvider` behaves in exactly the same way:

> Returns a std::string containing |num_bytes| of input data. Using this and
> |.c_str()| on the resulting string is the best way to get an immutable
> null-terminated C string. If fewer than |num_bytes| of data remain, returns
> a shorter std::string containing all of the data that's left.

https://github.com/llvm/llvm-project/blob/0fc8fc75a9d505dde92c0d6d8826b6384dfb5e2a/compiler-rt/include/fuzzer/FuzzedDataProvider.h#L133C1-L136C65

CMakeLists.txt
tyler92: @hartwork I'm not sure I understood your question, could you please clarify it?

tyler92:
> Please cherry-pick https://github.com/uriparser/uriparser/commit/e62e6ea1a7c8ea8ae09be6f69ad4c60d7c2957d1.

Done

> They only changed needed more on the side of your commits would be this: https://github.com/uriparser/uriparser/commit/4998d724155e504db681a0d9fb6c7a896babd496. if you squash that into the right commit, that would be perfect.

I applied this change as part of the review feedback https://github.com/uriparser/uriparser/pull/209#discussion_r1903329511

CMakeLists.txt
hartwork: @tyler92 the mix of `-fsanitize=fuzzer-no-link` and `-fsanitize=fuzzer` for the fuzzing CLI extecutables does not yet make sense to me. Why is it not 2x `-fsanitize=fuzzer`? https://llvm.org/docs/LibFuzzer.html#id24 would imply that for our case. I guess both work but one is better fitting our semantics, or I missing something?
I just want to understand it and then potentially copy the new version over to Expat.

hartwork:
> @hartwork I've corrected commit messages

@tyler92 it seems to say "formatiing" rather than "formatting". 

> > Please cherry-pick [e62e6ea](https://github.com/uriparser/uriparser/commit/e62e6ea1a7c8ea8ae09be6f69ad4c60d7c2957d1).
> 
> Done

Thanks!

> > They only changed needed more on the side of your commits would be this: [4998d72](https://github.com/uriparser/uriparser/commit/4998d724155e504db681a0d9fb6c7a896babd496). if you squash that into the right commit, that would be perfect.
> 
> I applied this change as part of the review feedback [#209 (comment)](https://github.com/uriparser/uriparser/pull/209#discussion_r1903329511)

Perfect!

hartwork (changes requested):
@tyler92 commit "Improve fuzzing targets by covering Unicode functions" seems to be doing mutliple things:
- Extend fuzzing to wide string functions
- Starting to use "chars_required + 1" instead of "chars_required"
- Extend the CMake build system to build fuzzers

Would it be too much to ask to split it up to three commits, ideally in that order?

That and the "formatiing" typo are not unlikely to be the last thing to ask from you in this pull request. Good work and thanks for all the timely adjustments! :pray:

CMakeLists.txt
```diff
@@ -334,6 +336,73 @@ if(URIPARSER_BUILD_TESTS)
...
+
+    function(add_fuzzer target_name source compile_definitions)
+        add_executable(${target_name} ${source})
+        target_link_libraries(${target_name} PRIVATE fuzzparser)
+        target_compile_definitions(${target_name} PRIVATE ${compile_definitions})
+        if(URIPARSER_OSSFUZZ_BUILD)
+            set_target_properties(${target_name} PROPERTIES LINK_FLAGS $ENV{LIB_FUZZING_ENGINE})
+            set_target_properties(${target_name} PROPERTIES LINKER_LANGUAGE "CXX")
+        else()
+            target_compile_options(${target_name} PRIVATE -fsanitize=fuzzer-no-link)
```
hartwork, line 378:
```suggestion
            target_compile_options(${target_name} PRIVATE -fsanitize=fuzzer)
```

?

Related to https://github.com/uriparser/uriparser/pull/209#discussion_r1903331297

tyler92: I checked it - you are right, it works well. Applied your suggestion

tyler92:
> @tyler92 commit "Improve fuzzing targets by covering Unicode functions" seems to be doing mutliple things:
> 
>     * Extend fuzzing to wide string functions
> 
>     * Starting to use "chars_required + 1" instead of "chars_required"
> 
>     * Extend the CMake build system to build fuzzers
> 
> 
> Would it be too much to ask to split it up to three commits, ideally in that order?
> 
> That and the "formatiing" typo are not unlikely to be the last thing to ask from you in this pull request. Good work and thanks for all the timely adjustments! 🙏

Done. Sorry for the typos in commit messages, there are too many rebases :slightly_smiling_face:

hartwork (approved):
@tyler92 thanks for this, good stuff! :+1:

CMakeLists.txt
hartwork: For anyone else reading: we changed it to `-fsanitize=fuzzer` in the meantime.

hartwork:
@tyler92 I've taking a few smaller learnings from this upstream to Expat now at https://github.com/libexpat/libexpat/pull/942 . If I missed anything or anything looks off, review is welcome (but please don't feel pushed to).

hartwork:
@tyler92 PS: please feel free to "CC @hartwork" me on upcoming follow-up pull-requests at https://github.com/google/oss-fuzz/pulls .

PR #212, opened 2025-01-06
`fuzzing.yml`: Start fuzzing wide string corpora, too

hartwork:
CC @tyler92

### DIFF
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -91,6 +91,7 @@ jobs:
         set -x -o pipefail
         cd "build/corpus_${fuzzer_name}/"
         find . -type f | sort | xargs -n 1000 "../fuzz/${fuzzer_name}" "${fuzz_args[@]}"
+        find . -type f | wc -l
 
     - name: Store fuzzing logs of last batch
       if: always()
--- a/.github/workflows/fuzzing.yml
+++ b/.github/workflows/fuzzing.yml
@@ -15,8 +15,20 @@ permissions:
 
 jobs:
   run_fuzzers:
-    name: Run fuzzing regression tests
+    name: ${{ matrix.fuzzer_name }}
+    strategy:
+      fail-fast: false
+      matrix:
+        fuzzer_name:
+          - uri_dissect_query_malloc_fuzzer
+          - uri_dissect_query_mallocw_fuzzer
+          - uri_free_fuzzer
+          - uri_freew_fuzzer
+          - uri_parse_fuzzer
+          - uri_parsew_fuzzer
     runs-on: ubuntu-24.04
+    env:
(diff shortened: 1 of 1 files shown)


tyler92:
LGTM but:

1. Public corpus for new targets will be created not earlier than tomorrow (today's run doesn't include new fuzz targets)
2. It seems there has been an error on OSS-Fuzz (?) side last 4 days: https://oss-fuzz-build-logs.storage.googleapis.com/index.html#uriparser (open "Coverage Builds" tab) related to corpus. I don't know if it affects this PR or not but I see several projects with this error without an explanation what's wrong

hartwork:
@tyler92 any idea why this still doesn't work today?

tyler92:
> @tyler92 any idea why this still doesn't work today?

Unfortunately no. I see that fuzzing for new targets works, e.g.: https://storage.googleapis.com/oss-fuzz-coverage/uriparser/reports-by-target/20250112/uri_parsew_fuzzer/linux/report.html

but meantime there are issues with the corpus: https://oss-fuzz-build-logs.storage.googleapis.com/log-57e31033-f954-40e3-9d1b-405b402db19b.txt

```
Step #5: [/corpus/uri_free_fuzzer.zip]
Step #5:   End-of-central-directory signature not found.  Either this file is not
Step #5:   a zipfile, or it constitutes one disk of a multi-part archive.  In the
Step #5:   latter case the central directory and zipfile comment will be found on
Step #5:   the last disk(s) of this archive.
Step #5: unzip:  cannot find zipfile directory in one of /corpus/uri_free_fuzzer.zip or
Step #5:         /corpus/uri_free_fuzzer.zip.zip, and cannot find /corpus/uri_free_fuzzer.zip.ZIP, period.
Step #5: Failed to unpack the corpus for uri_free_fuzzer. This usually means that corpus backup for a particular fuzz target does not exist. If a fuzz target was added in the last 24 hours, please wait one more day. Otherwise, something is wrong with the fuzz target or the infrastructure, and corpus pruning task does not finish successfully.
```

I see this issue for other projects too, so I _assume_ this is something on the OSS-Fuzz side. Probably you as a maintainer can ask [here](https://github.com/google/oss-fuzz/issues) what's wrong (I already asked for other projects but there is no response yet).

hartwork:
@tyler92 interesting, thanks!

PS: I wish they would wake up about https://github.com/google/oss-fuzz/pull/10504 also but I sort of gave up.

hartwork:
Note to self: https://github.com/google/oss-fuzz/issues/12948#issuecomment-2741595921

hartwork:
More notes to self:
- https://github.com/google/oss-fuzz/issues/12948#issuecomment-2782588537
- https://github.com/google/oss-fuzz/issues/13195

PR #222, opened 2025-04-02
Normalized percent-encoded octets should have uppercase letters in the host (fixes #221)

### LINKED ISSUE #221: Normalized percent-encoded octets should have uppercase letters in the host
One of my colleagues spotted a small bug in my URI parser implementation that uses uriparser. 

Digging into the the code, I realized that this happens because the lowercasing normalization of the host is performed after percent-encoding normalization, effectively partly reverting the result of the former normalization.

An example:

`https://%e4%bd%a0%e5%a5%bd%e4%bd%a0%e5%a5%bd.com` ([https://你好你好.com](https://xn--6qqa088eba.com/)) should be normalized to `[https://%E4%BD%A0%E5%A5%BD%E4%BD%A0%E5%A5%BD.com](https://xn--6qqa088eba.com/)`.

I have a simple - and hopefully correct - patch, so I'll file it very soon.

kocsismate:
A simple solution that skips lowercasing 2 characters after each `%`.

Fixes #221

### DIFF
--- a/src/UriNormalize.c
+++ b/src/UriNormalize.c
@@ -636,9 +636,8 @@ static URI_INLINE int URI_FUNC(NormalizeSyntaxEngine)(URI_TYPE(Uri) * uri,
 				uri->hostText.first = uri->hostData.ipFuture.first;
 				uri->hostText.afterLast = uri->hostData.ipFuture.afterLast;
 			} else if ((uri->hostText.first != NULL)
-					&& (uri->hostData.ip4 == NULL)
-					&& (uri->hostData.ip6 == NULL)) {
-				/* Regname */
+					&& (uri->hostData.ip4 == NULL)) {
+				/* Regname or IPv6 */
 				if (uri->owner) {
 					URI_FUNC(FixPercentEncodingInplace)(uri->hostText.first,
 							&(uri->hostText.afterLast));
--- a/src/UriNormalize.c
+++ b/src/UriNormalize.c
@@ -104,6 +104,8 @@ static UriBool URI_FUNC(ContainsUglyPercentEncoding)(const URI_CHAR * first,
 
 static void URI_FUNC(LowercaseInplace)(const URI_CHAR * first,
 		const URI_CHAR * afterLast);
+static void URI_FUNC(LowercaseInplaceExceptPercentEncoding)(const URI_CHAR * first,
+		const URI_CHAR * afterLast);
 static UriBool URI_FUNC(LowercaseMalloc)(const URI_CHAR ** first,
 		const URI_CHAR ** afterLast, UriMemoryManager * memory);
 
@@ -238,6 +240,26 @@ static URI_INLINE void URI_FUNC(LowercaseInplace)(const URI_CHAR * first,
 		for (; i < afterLast; i++) {
 			if ((*i >= _UT('A')) && (*i <=_UT('Z'))) {
 				*i = (URI_CHAR)(*i + lowerUpperDiff);
+			}
+		}
+	}
+}
+
+
+
+static URI_INLINE void URI_FUNC(LowercaseInplaceExceptPercentEncoding)(const URI_CHAR * first,
+		const URI_CHAR * afterLast) {
+	if ((first != NULL) && (afterLast != NULL) && (afterLast > first)) {
+		URI_CHAR * i = (URI_CHAR *)first;
+		const int lowerUpperDiff = (_UT('a') - _UT('A'));
+		for (; i < afterLast; i++) {
+			if ((*i >= _UT('A')) && (*i <=_UT('Z'))) {
+				*i = (URI_CHAR)(*i + lowerUpperDiff);
+			} else if (*i == _UT('%')) {
+				if (i + 3 >= afterLast) {
+					return;
+				}
+				i += 2;
 			}
 		}
 	}
@@ -631,7 +653,7 @@ static URI_INLINE int URI_FUNC(NormalizeSyntaxEngine)(URI_TYPE(Uri) * uri,
 					doneMask |= URI_NORMALIZE_HOST;
 				}
 
-				URI_FUNC(LowercaseInplace)(uri->hostText.first,
+				URI_FUNC(LowercaseInplaceExceptPercentEncoding)(uri->hostText.first,
 						uri->hostText.afterLast);
 			}
 		}
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -1460,6 +1460,10 @@ TEST(UriSuite, TestNormalizeSyntax) {
 				L"https://%E4%BD%A0%E5%A5%BD%E4%BD%A0%E5%A5%BD.com"));
 
 		ASSERT_TRUE(testNormalizeSyntaxHelper(
+				L"https://[2041:0000:140F::875B:131B]",
+				L"https://[2041:0000:140f::875b:131b]"));
+
+		ASSERT_TRUE(testNormalizeSyntaxHelper(
 				L"HTTP://a:b@HOST:123/./1/2/../%41?abc#def",
 				L"http://a:b@host:123/1/A?abc#def"));
 
--- a/test/test.cpp
+++ b/test/test.cpp
@@ -1456,6 +1456,10 @@ TEST(UriSuite, TestNormalizeSyntax) {
 				L"http://user:pass@somehost.com:123"));
 
 		ASSERT_TRUE(testNormalizeSyntaxHelper(
+				L"https://%e4%bd%a0%e5%a5%bd%e4%bd%a0%e5%a5%bd.com",
+				L"https://%E4%BD%A0%E5%A5%BD%E4%BD%A0%E5%A5%BD.com"));
+
+		ASSERT_TRUE(testNormalizeSyntaxHelper(
 				L"HTTP://a:b@HOST:123/./1/2/../%41?abc#def",
 				L"http://a:b@host:123/1/A?abc#def"));
 


src/UriNormalize.c
```diff
@@ -245,6 +247,24 @@ static URI_INLINE void URI_FUNC(LowercaseInplace)(const URI_CHAR * first,
...
+		const URI_CHAR * afterLast) {
+	if ((first != NULL) && (afterLast != NULL) && (afterLast > first)) {
+		URI_CHAR * i = (URI_CHAR *)first;
+		const int lowerUpperDiff = (_UT('a') - _UT('A'));
+		URI_CHAR * skipLowercasingUntilPosition = NULL;
+		for (; i < afterLast; i++) {
+			if ((*i >= _UT('A')) && (*i <=_UT('Z')) && skipLowercasingUntilPosition < i) {
+				*i = (URI_CHAR)(*i + lowerUpperDiff);
+			} else if (*i == _UT('%')) {
+				skipLowercasingUntilPosition = i + 2;
```
kocsismate, line 260: I'm not deeply familiar with percent-encodings... So I don't know if it's possible that there is a `%` after another `%`. If it's not (and 2 characters after a `%` sign can always be skipped), then I believe only `i` could be incremented by 3.

hartwork: @kocsismate I believe I found part of the answer to that here:

> [2.4](https://datatracker.ietf.org/doc/html/rfc3986#section-2.4).  When to Encode or Decode
> [..]
> Because the percent ("%") character serves as the indicator for percent-encoded octets, it must be percent-encoded as "%25" for that octet to be used as data within a URI.

hartwork (changes requested):
@kocsismate I only had a first quick look, some thoughts:

The current style is not close to mine and I would like to offer you a swap of seats where you get to be the one doing review and I take the coding seat so that we get to a result matching my handwriting without a lot of ping-ping and less work on both sides. Would that work for you? We'll need to add additional tests also, and writing them would then be on me.

We'll need to check what the other place calling `LowercaseInplace` needs to do, with regard to RFC 3986 (have not yet checked):

```console
# git grep -n -A1 LowercaseInplace | tail -n 5
src/UriNormalize.c:604:                                 URI_FUNC(LowercaseInplace)(uri->hostData.ipFuture.first,
src/UriNormalize.c-605-                                                 uri->hostData.ipFuture.afterLast);
--
src/UriNormalize.c:634:                         URI_FUNC(LowercaseInplace)(uri->hostText.first,
src/UriNormalize.c-635-                                         uri->hostText.afterLast);
```

We'll need to check the other place that produces lowercase hex digits at the moment:

```console
# git --no-pager grep -n 'HexToLetter.*URI_FALSE'
src/UriRecompose.c:276:                                                                 text[0] = URI_FUNC(HexToLetterEx)(value / 16, URI_FALSE);
src/UriRecompose.c:277:                                                                 text[1] = URI_FUNC(HexToLetterEx)(value % 16, URI_FALSE);
```

Both are related to IPvFuture.

What do you think?

kocsismate:
I'm currently off for a while, but I'll check the other usages, and change the code with regards to percent-encoding of `%`.

Could you please approve the GitHub workflow so that CI can run?

> The current style is not close to mine (of today, the code is old)

I'm fine to follow any other coding style if you can link me some piece of code that is written accoridng to the current style.

kocsismate:
> We'll need to check what the other place calling LowercaseInplace needs to do, with regard to RFC 3986 (have not yet checked):

According to the spec, hexadecimal digits inside the host should be lowercase:

> Although host is case-insensitive, producers and normalizers should use lowercase for registered names and hexadecimal addresses for the sake of uniformity, while only using uppercase letters for percent-encodings.

The other usage is within the scheme that is also fine as-is.

hartwork:
@kocsismate thanks for the updates and your replies! I need a few more days to get back to you on this — see you soon!

kocsismate:
> According to the spec, hexadecimal digits inside the host should be lowercase:

I only noticed a few days ago that this normalization hadn't been implemented until now. It seemed to me that it was already implemented, but then I realized that the condition I changed used to exclude ipv6 hosts, so I took care of this too in my last commit.

src/UriNormalize.c
```diff
@@ -245,6 +247,23 @@ static URI_INLINE void URI_FUNC(LowercaseInplace)(const URI_CHAR * first,
...
+		const URI_CHAR * afterLast) {
+	if ((first != NULL) && (afterLast != NULL) && (afterLast > first)) {
+		URI_CHAR * i = (URI_CHAR *)first;
+		const int lowerUpperDiff = (_UT('a') - _UT('A'));
+		for (; i < afterLast; i++) {
+			if ((*i >= _UT('A')) && (*i <=_UT('Z'))) {
+				*i = (URI_CHAR)(*i + lowerUpperDiff);
+			} else if (*i == _UT('%')) {
+				i += 2;
+			}
```
hartwork, line 260: While this seems safe with well-formed data, it could advance the pointer past the end of the array for malformed input. I would rather be safe then sorry. How about:

```suggestion
			} else if (*i == _UT('%')) {
				if (i + 3 >= afterLast) {
					return;
				}
				i += 2;
			}
```

src/UriNormalize.c
```diff
@@ -245,6 +247,23 @@ static URI_INLINE void URI_FUNC(LowercaseInplace)(const URI_CHAR * first,
 
 
 
+static URI_INLINE void URI_FUNC(LowercaseInplaceAfterPercentEncoding)(const URI_CHAR * first,
```
hartwork, line 250: Could you name the function `LowercaseInplaceExceptPercentEncoding` with "Except" rather than "After"? It's close but takes takes time out of the picture.

test/FourSuite.cpp
```diff
@@ -604,6 +604,8 @@ bool normalizeAndCompare(const char * uriText,
 
 TEST(FourSuite, CaseNormalizationTests) {
 	ASSERT_TRUE(normalizeAndCompare("HTTP://www.EXAMPLE.com/", "http://www.example.com/"));
+	ASSERT_TRUE(normalizeAndCompare("https://%e4%bd%a0%e5%a5%bd%e4%bd%a0%e5%a5%bd.com", "https://%E4%BD%A0%E5%A5%BD%E4%BD%A0%E5%A5%BD.com"));
+	ASSERT_TRUE(normalizeAndCompare("https://[2041:0000:140F::875B:131B]", "https://[2041:0000:140f::875b:131b"));
```
hartwork, line 608: - I like the general approach
- The second test is missing a trailing `]` (closing square bracket).
- Section `FourSuite` of the test suite relates to [the 4Suite project](https://sourceforge.net/projects/foursuite/) so these two new tests need a different place — in file `test/test.cpp` near or in `TEST(UriSuite, TestNormalizeSyntax) {` would be a good candidate for a new location.

hartwork (changes requested):
Hi @kocsismate, I found time for review again now.  I like the new version better, thanks! :+1:

Please check the line based comments below: It's all details, this is not far from mergability.

Also please rebase into latest `master` (e.g. for recent extension of the CI in #212).

While rebasing please squash the two commits.  I'm not objecting to multiple commits or a separate commit for tests and implementation but commit `Optimize code + add tests` is [two different things in a single commit](https://blog.hartwork.org/posts/my-approach-to-code-review/#C9) and the optimize part could be squashed into the former commit for cleaner history. :pray:

With regard to my earlier questions from https://github.com/uriparser/uriparser/pull/222#pullrequestreview-2741713743 and your answers I have re-checked these now and found (if of interest):

- There is `LowercaseInplace` and also `LowercaseMalloc`.

- `LowercaseInplace` is used in three places:
  - normalization of `scheme` — [cannot contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), already good
  - normalization of `IPvFuture` host — [cannot contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), already good
  - normalization of `reg-name` host — [**can** contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), needs a fix, done in this pull request, great :+1:

- `LowercaseMalloc` is used in two places:
  - normalization of `scheme` — [cannot contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), already good
  - normalization of `IPvFuture` host — [cannot contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), already good
  
- The only place using `HexToLetter(Ex)` with lowercase is in `ToStringEngine` — already good.

I'm looking forward to the next iteration :beers:

kocsismate:
Hi,

Thank you for the review :)

> Also please rebase into latest master (e.g. for recent extension of the CI in https://github.com/uriparser/uriparser/pull/212).

Done.

> While rebasing please squash the two commits.

Until the review is done, I prefer to just push the changes so that the reviewer can easily review only the new changes, and I squash (or rewrite the git history meaningfully) before merge. Does this approach work for you? Now, I'm about to push another fix for the review comments, and I'm wondering what is the most convenient for you.

> normalization of reg-name host — [can contain percent encoding](https://datatracker.ietf.org/doc/html/rfc3986#appendix-A), needs a fix, done in this pull request, great 👍

Yes, I didn't realize until now that I forgot to push the fix for a couple of weeks.

src/UriNormalize.c
```diff
@@ -614,9 +636,8 @@ static URI_INLINE int URI_FUNC(NormalizeSyntaxEngine)(URI_TYPE(Uri) * uri,
 				uri->hostText.first = uri->hostData.ipFuture.first;
 				uri->hostText.afterLast = uri->hostData.ipFuture.afterLast;
 			} else if ((uri->hostText.first != NULL)
-					&& (uri->hostData.ip4 == NULL)
-					&& (uri->hostData.ip6 == NULL)) {
```
kocsismate, line 618: Having this line made the ipv6 tests fail because it skipped lowercasing normalization

hartwork:
> During reviews, I usually prefer to just push my changes for each review round so that the reviewer can easily review only the new changes, and when everything seems OK, I squash (or rewrite the git history meaningfully) before merge. Does this approach work for you?

@kocsismate no worries, my main concern is the commits at merge time, please do as you see fit in the time before that :+1:

src/UriNormalize.c
hartwork: @kocsismate okay, but from a quick first look it doesn't look like the right fix. I'll look more closely soon and report back.

hartwork: @kocsismate update: I re-checked: for IPv6 addresses both `.hostText` and `hostData.ip6` are set — you're right, good fix! :+1:

hartwork (changes requested):
@kocsismate I believe we're already at the point where code and CI are already great and nothing but re-cutting commits remains. I'd be good with all in one commit or one commit fix and one commit tests. I'm also open to other ideas.

kocsismate:
@hartwork I pushed my changes in a single commit. If you want, I can split it into two commits, because the fix for the lowercase normalization of IPv6 hosts is effectively a separate one. I mentioned the additional change in the commit description for any case.

hartwork (changes requested):
@kocsismate I checked the commit and its message and have a more clear picture now.

The commit message says "n the same time" with "n" instead of "in" (or potentially better "at", but I could be wrong) which is why we should touch the commit again in any case.

I see the two parts more clearly now.  How about this cutting:

- The first commit…
  - Adds and uses new function `LowercaseInplaceExceptPercentEncoding`
  - Adds the regname test `%e4%bd%a0%e5%a5%bd%e4%bd%a0%e5%a5%bd.com`
  - Has a commit message a la `UriNormalize.c: Percent-encoding should use uppercase letters\n\n.. when normalized according to RFC 3986 section 6.2.2.1; fixes issue #221.`

- The second commit…
  - Does the `/* Regname or IPv6 */` adjustment
  - Adds the IPv6 test
  - Has a commit message a la `UriNormalize.c: Hex values in IPv6 hosts should use lowercase letters\n\n.. when normalized according to RFC 3986 section 3.2.2.`
  
What do you think?

kocsismate:
> What do you think?

Yes, doing exactly this was my other plan :) So I'm going to change my commits again.

P.S. I'm also going to replay to your message in the other ticket in the coming days.

hartwork (approved):
> > What do you think?
> 
> Yes, doing exactly this was my other plan :) So I'm going to change my commits again.

@kocsismate perfect, thanks for the adjustments! :+1: 

> P.S. I'm also going to reply to your message in the other ticket in the coming days.

Sounds good :+1:

PR #230, opened 2025-06-16
Fix #200 Add support for copying Uri structures

### LINKED ISSUE #200: Duplicating URIs?
Hi!
I wish to use this library in a C++ project, by crafting a wrapper around it to make it more memory safe, however, while poking around, I couldn't find a good way of duplicating `UriUriA` objects. What is the proper way of performing a deep copy of `UriUriA`s?

A simple shallow clone, `owner=false`, `uriMakeOwnerA(&uri)` didn't do the trick and later member releases performed a double free.

kocsismate:
I also need this functionality for https://wiki.php.net/rfc/url_parsing_api because of multiple reasons:

- In order to maintain immutability of URIs, the structs have to be copied first before any modification
- The PHP `Uri` class exposes both the normalized and non-normalized URI components (therefore it internally stores two Uri structs). When a normalized component is requested, normalization is done on demand, by copying the original struct first

The patch needs a last few touches, including tests, but I wanted to propose it nevertheless to get early feedback.

### DIFF
--- a/src/UriCommon.c
+++ b/src/UriCommon.c
@@ -119,6 +119,23 @@ int URI_FUNC(CompareRange)(
 
 
 
+UriBool URI_FUNC(CopyRange)(URI_TYPE(TextRange) * destRange,
+		const URI_TYPE(TextRange) * sourceRange, UriMemoryManager * memory) {
+	const int lenInChars = (int)(sourceRange->afterLast - sourceRange->first);
+	const int lenInBytes = lenInChars * sizeof(URI_CHAR);
+	URI_CHAR * dup = memory->malloc(memory, lenInBytes);
+	if (dup == NULL) {
+		return URI_FALSE;
+	}
+	memcpy(dup, sourceRange->first, lenInBytes);
+	destRange->first = dup;
+	destRange->afterLast = dup + lenInChars;
+
+	return URI_TRUE;
+}
+
+
+
 UriBool URI_FUNC(RemoveDotSegmentsEx)(URI_TYPE(Uri) * uri,
 		UriBool relative, UriBool pathOwned, UriMemoryManager * memory) {
 	URI_TYPE(PathSegment) * walker;
--- a/src/UriCommon.c
+++ b/src/UriCommon.c
@@ -136,6 +136,23 @@ UriBool URI_FUNC(CopyRange)(URI_TYPE(TextRange) * destRange,
 
 
 
+UriBool URI_FUNC(CopyRangeAsNeeded)(URI_TYPE(TextRange) * destRange,
+		const URI_TYPE(TextRange) * sourceRange, UriBool useSafe, UriMemoryManager * memory) {
+	if (sourceRange->first == NULL) {
+		destRange->first = NULL;
+		destRange->afterLast = NULL;
+	} else if (sourceRange->first == sourceRange->afterLast && useSafe) {
+		destRange->first = URI_FUNC(SafeToPointTo);
+		destRange->afterLast = URI_FUNC(SafeToPointTo);
+	} else {
+		return URI_FUNC(CopyRange)(destRange, sourceRange, memory);
+	}
+
+	return URI_TRUE;
+}
+
+
+
 UriBool URI_FUNC(RemoveDotSegmentsEx)(URI_TYPE(Uri) * uri,
 		UriBool relative, UriBool pathOwned, UriMemoryManager * memory) {
 	URI_TYPE(PathSegment) * walker;
--- a/src/UriCopy.c
+++ b/src/UriCopy.c
@@ -1,0 +1,234 @@
+/*
+ * uriparser - RFC 3986 URI parsing library
+ *
+ * Copyright (C) 2007, Weijia Song <songweijia@gmail.com>
+ * Copyright (C) 2007, Sebastian Pipping <sebastian@pipping.org>
+ * Copyright (C) 2025, Máté Kocsis <kocsismate@php.net>
+ * All rights reserved.
+ *
+ * Redistribution and use in source  and binary forms, with or without
+ * modification, are permitted provided  that the following conditions
+ * are met:
+ *
+ *     1. Redistributions  of  source  code   must  retain  the  above
+ *        copyright notice, this list  of conditions and the following
+ *        disclaimer.
+ *
+ *     2. Redistributions  in binary  form  must  reproduce the  above
+ *        copyright notice, this list  of conditions and the following
+ *        disclaimer  in  the  documentation  and/or  other  materials
+ *        provided with the distribution.
+ *
+ *     3. Neither the  name of the  copyright holder nor the  names of
+ *        its contributors may be used  to endorse or promote products
+ *        derived from  this software  without specific  prior written
+ *        permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND  ANY EXPRESS OR IMPLIED WARRANTIES,  INCLUDING, BUT NOT
+ * LIMITED TO,  THE IMPLIED WARRANTIES OF  MERCHANTABILITY AND FITNESS
+ * FOR  A  PARTICULAR  PURPOSE  ARE  DISCLAIMED.  IN  NO  EVENT  SHALL
+ * THE  COPYRIGHT HOLDER  OR CONTRIBUTORS  BE LIABLE  FOR ANY  DIRECT,
+ * INDIRECT, INCIDENTAL, SPECIAL,  EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+ * (INCLUDING, BUT NOT LIMITED TO,  PROCUREMENT OF SUBSTITUTE GOODS OR
+ * SERVICES; LOSS OF USE, DATA,  OR PROFITS; OR BUSINESS INTERRUPTION)
+ * HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
+ * STRICT  LIABILITY,  OR  TORT (INCLUDING  NEGLIGENCE  OR  OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
+ * OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file UriCopy.c
+ * Holds the RFC 3986 %URI normalization implementation.
+ * NOTE: This source file includes itself twice.
+ */
+
+/* What encodings are enabled? */
+#include <uriparser/UriDefsConfig.h>
+#if (!defined(URI_PASS_ANSI) && !defined(URI_PASS_UNICODE))
+/* Include SELF twice */
+# ifdef URI_ENABLE_ANSI
+#  define URI_PASS_ANSI 1
+#  include "UriCopy.c"
+#  undef URI_PASS_ANSI
+# endif
+# ifdef URI_ENABLE_UNICODE
+#  define URI_PASS_UNICODE 1
+#  include "UriCopy.c"
+#  undef URI_PASS_UNICODE
+# endif
+#else
+# ifdef URI_PASS_ANSI
+#  include <uriparser/UriDefsAnsi.h>
+# else
+#  include <uriparser/UriDefsUnicode.h>
+#  include <wchar.h>
+# endif
+
+
+
+#ifndef URI_DOXYGEN
+# include <uriparser/Uri.h>
+# include "UriCommon.h"
+# include "UriMemory.h"
+# include "UriNormalize.h"
+# include "UriCopy.h"
+#endif
+
+
+
+static void URI_FUNC(PreventLeakageAfterCopy)(URI_TYPE(Uri) * uri,
+		unsigned int revertMask, UriMemoryManager * memory) {
+	URI_FUNC(PreventLeakage)(uri, revertMask, memory);
+
+	if (uri->hostData.ip4 != NULL) {
+		memory->free(memory, uri->hostData.ip4);
+		uri->hostData.ip4 = NULL;
+	} else if (uri->hostData.ip6 != NULL) {
+		memory->free(memory, uri->hostData.ip6);
+		uri->hostData.ip6 = NULL;
+	}
+
+	if (revertMask & URI_NORMALIZE_PORT) {
+		if (uri->portText.first != uri->portText.afterLast) {
+			memory->free(memory, (URI_CHAR *)uri->portText.first);
+		}
+		uri->portText.first = NULL;
+		uri->portText.afterLast = NULL;
+	}
+}
+
+
+
+int URI_FUNC(CopyUriMm)(URI_TYPE(Uri) * destUri,
+		 const URI_TYPE(Uri) * sourceUri, UriMemoryManager * memory) {
+	unsigned int doneMask = URI_NORMALIZED;
+
+	if (sourceUri == NULL || destUri == NULL) {
+		return URI_ERROR_NULL;
+	}
+
+	URI_CHECK_MEMORY_MANAGER(memory); /* may return */
+
+	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->scheme, &sourceUri->scheme, URI_FALSE, memory) == URI_FALSE) {
+		return URI_ERROR_MALLOC;
+	}
+
+	doneMask |= URI_NORMALIZE_SCHEME;
+
+	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->userInfo, &sourceUri->userInfo, URI_FALSE, memory) == URI_FALSE) {
+		URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+		return URI_ERROR_MALLOC;
+	}
+
+	doneMask |= URI_NORMALIZE_USER_INFO;
+
+	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->hostText, &sourceUri->hostText, URI_TRUE, memory) == URI_FALSE) {
+		URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+		return URI_ERROR_MALLOC;
+	}
+
+	doneMask |= URI_NORMALIZE_HOST;
+
+	if (sourceUri->hostData.ip4 == NULL) {
+		destUri->hostData.ip4 = NULL;
+	} else {
+		destUri->hostData.ip4 = memory->malloc(memory, sizeof(UriIp4));
+		if (destUri->hostData.ip4 == NULL) {
+			URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+			return URI_ERROR_MALLOC;
+		}
+		*(destUri->hostData.ip4) = *(sourceUri->hostData.ip4);
+	}
+
+	if (sourceUri->hostData.ip6 == NULL) {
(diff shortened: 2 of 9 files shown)


hartwork (changes requested):
@kocsismate I'm only starting to have a closer look. From what I see so far:

- the pull request [does multiple things in a single commit](https://blog.hartwork.org/posts/my-approach-to-code-review/#C9): e.g. move code around and add new code — that's too much for a single commit and the move itself does not seem strictly necessary to me as exposing needed functions through their related `.h` files would also make them available for use and then the move is fully optional (and something I would want to do as a potential refactoring myself later to keep this pull request to the point). I would ask to revert/drop all moves and expose through existing headers instead, and to add the new copy code to new files `UriCopy.{c,h}`, and to then get back to detail level review. If some move is more necessary than I understood, it should be necessary moves only for now and in a dedicated commit please. Thank you!
- The way that function `PreventLeakage` is used looks mistaken: the `revertMask` input would need to grow with each allocation a la `doneMask |= URI_NORMALIZE_SCHEME;` — `git grep -F '|=' -- src/UriNormalize.c` yields examples of that.

I'll pause review here for now, waiting for feedback.

kocsismate:
> I'll pause review here for now, waiting for feedback.

Thanks for the input! I'll try to address your suggestion ASAP, possibly around Wednesday!

src/UriCommon.c
```diff
@@ -119,6 +120,225 @@ int URI_FUNC(CompareRange)(
...
+	} else {
+		return URI_FUNC(CopyRangeEngine)(destRange, sourceRange, memory);
+	}
+
+	return URI_TRUE;
+}
+
+
+
+void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
```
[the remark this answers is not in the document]
kocsismate: > I don't see the need, please help me see it.

I only said this because including the `UriNormalize` header didn't entirely fit semantically. But I've just realized that I can simply add the function declaration to Uri.h. Sorry for the noise!

> I would ask to revert/drop all moves and expose through existing headers instead, and to add the new copy code to new files UriCopy.{c,h}, and to then get back to detail level review. 

Alright, I got it!

> My vote for using those despite the _NORMALIZE_ part for now. 

That's perfectly fine for me. I just wanted to know your opinion in the question, as I wasn't sure if the semantical discrepancy is something that you are willing to accept.

> Would you consider normalization like that a bug or a feature (and why)?

IMO it would be a feature. It's not mandated by the specification, therefore it's an optional - but pretty much logical - step to do. I would vote for adding the `URI_NORMALIZE_PORT` then in this PR for the purpose of copying, and a followup PR can make use of it during normalization.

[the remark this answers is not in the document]
kocsismate: > I'm not sure I understand 100% already: I would not expect any additions of functions to Uri.h other than CopyUriMm and CopyUri

Yes, I see it. In fact, after writing my comment, I had ended up implementing something similar that you asked.

>  I can take over that first half of port normalization to parallelize things if you want. Please let me know either way.

I would appreciate if you could take care of the port normalization. Most probably all my available time will be consumed by the implementation of my RFC + the setter support in uriparser. :(

P.S. Please don't forget to answer to our email.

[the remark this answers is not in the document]
hartwork: @kocsismate update: I have a first draft of port normalization plus tests at #231 now. Review welcome.

src/UriNormalize.c
```diff
@@ -109,12 +110,9 @@ static void URI_FUNC(LowercaseInplaceExceptPercentEncoding)(const URI_CHAR * fir
 static UriBool URI_FUNC(LowercaseMalloc)(const URI_CHAR ** first,
 		const URI_CHAR ** afterLast, UriMemoryManager * memory);
 
-static void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
-		unsigned int revertMask, UriMemoryManager * memory);
 
 
-
-static URI_INLINE void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
+void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
```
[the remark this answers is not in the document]
kocsismate: > Not with regard to normalization. Unless I am missing something. What makes you think it needs fixing?

I had the impression that `PreventLeakage` would destroy the whole structure, not just the parts that normalization may have touched. But I think I see it now more clearly: the stuffs that were not touched don't need to be freed, since `preventLeakage` only runs when the URI is not owned, and therefore there's nothing else to free. And sorry for the mess ups!

src/UriNormalize.c
```diff
@@ -152,7 +150,13 @@ static URI_INLINE void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
...
 	}
 
-	/* NOTE: Port cannot happen! */
+	if (revertMask & URI_NORMALIZE_PORT) {
+		if (uri->portText.first != uri->portText.afterLast) {
+			memory->free(memory, (URI_CHAR *)uri->portText.first);
+		}
+		uri->portText.first = NULL;
+		uri->portText.afterLast = NULL;
+	}
```
hartwork, line 159: I would prefer to have this code go [into a new function](https://github.com/uriparser/uriparser/pull/230#discussion_r2158710056) that calls out to `PreventLeakage` and does additional things on top. This way we will have code in there that is dead to the eyes of normalization.

hartwork (changes requested):
@kocsismate flushing some comments from review and taking a break from review for new, more to come…

include/uriparser/UriBase.h
```diff
@@ -303,7 +304,7 @@ typedef enum UriResolutionOptionsEnum {
  * @since 0.9.0
  */
 URI_PUBLIC int uriCompleteMemoryManager(UriMemoryManager * memory,
-		UriMemoryManager * backend);
+										UriMemoryManager * backend);
```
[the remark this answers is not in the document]
kocsismate: Hmm, likely CLion autoformatted it somwehow, but yes, I'll certainly get rid of it!

src/UriCopy.h
```diff
@@ -0,0 +1,83 @@
+/*
+ * uriparser - RFC 3986 URI parsing library
+ *
+ * Copyright (C) 2007, Weijia Song <songweijia@gmail.com>
+ * Copyright (C) 2007, Sebastian Pipping <sebastian@pipping.org>
```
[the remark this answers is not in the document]
hartwork: > I'd prefer using my PHP address because that's what I also use for the headers in php-src.

@kocsismate that's good with me :+1:

src/UriCopy.c
```diff
@@ -0,0 +1,238 @@
...
+	}
+}
+
+
+
+int URI_FUNC(CopyUriMm)(URI_TYPE(Uri) * destUri,
+		 const URI_TYPE(Uri) * sourceUri, UriMemoryManager * memory) {
+	if (sourceUri == NULL) {
+		return URI_ERROR_NULL;
+	}
```
hartwork, line 107:
```suggestion
	if (sourceUri == NULL || destUri == NULL) {
		return URI_ERROR_NULL;
	}
```
?

src/UriCopy.c
```diff
@@ -0,0 +1,238 @@
...
+
+	URI_CHECK_MEMORY_MANAGER(memory); /* may return */
+
+	unsigned int doneMask = URI_NORMALIZED;
```
hartwork, line 111: This line^^ violates C89/C90:

```
[..]/src/UriCopy.c:111:9: error: ISO C90 forbids mixed declarations and code [-Werror=declaration-after-statement]
  111 |         unsigned int doneMask = URI_NORMALIZED;
      |         ^~~~~~~~
```

Please add files `UriCopy.{c,h}` to the build system like this…

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 344e5e7..a0d772f 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -144,6 +147,8 @@ set(LIBRARY_CODE_FILES
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCommon.c
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCommon.h
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCompare.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCopy.c
+    ${CMAKE_CURRENT_SOURCE_DIR}/src/UriCopy.h
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriEscape.c
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriFile.c
     ${CMAKE_CURRENT_SOURCE_DIR}/src/UriIp4Base.c
```

…and rebase onto latest `master` for merged #232. Local compilation will show errors then and CI will build the code (and come to the same results).

src/UriCopy.c
```diff
@@ -0,0 +1,238 @@
...
+	} else {
+		destUri->hostData.ip4 = memory->malloc(memory, sizeof(UriIp4));
+		if (destUri->hostData.ip4 == NULL) {
+			URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+			return URI_ERROR_MALLOC;
+		}
+		*(destUri->hostData.ip4) = *(sourceUri->hostData.ip4);
+	}
+
+	doneMask |= URI_NORMALIZE_HOST;
```
hartwork, line 144:
```suggestion
```

(because `URI_NORMALIZE_HOST` controls `.hostText` in `PreventLeakage` only and also it has been set 13 lines further up already)

src/UriCopy.h
```diff
@@ -0,0 +1,80 @@
...
+#  include <uriparser/UriDefsAnsi.h>
+# else
+#  define URI_COPY_H_UNICODE 1
+#  include <uriparser/UriDefsUnicode.h>
+# endif
+
+
+
+void URI_FUNC(PreventLeakage)(URI_TYPE(Uri) * uri,
+		unsigned int revertMask, UriMemoryManager * memory);
```
hartwork, line 73: This prototype should probably go to `src/UriNormalize.h` instead while the function is defined in `src/UriNormalize.c` — am I missing something?

hartwork (changes requested):
@kocsismate I have yet to review the big loop. Flushing some comments…

src/UriCopy.c
```diff
@@ -0,0 +1,238 @@
...
+			walkerNew = walkerNew->next;
+			if (URI_FUNC(CopyRangeAsNeeded)(&walkerNew->text, &walker->text, URI_TRUE, memory) == URI_FALSE) {
+				URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+				return URI_ERROR_MALLOC;
+			}
+			walkerNew->reserved = NULL;
+			walker = walker->next;
+		}
+		walkerNew->next = NULL;
+		destUri->pathTail = walkerNew;
```
hartwork, line 206: I find this branch of the code overly complex even before being bullet proof (see above, so more complexity coming). Let me share python-like pseudo code of how I would approach it:

```
if sourceUri->pathHead:
    destUri->pathHead = NULL
    destUri->pathTail = NULL

    sourceWalker = destUri->pathHead
    destPrev = NULL
    
    while sourceWalker:
        destWalker = MALLOC(..)
        
        if destWalker == NULL:
            prevent leak
            return ....

        destWalker->text.first = NULL
        destWalker->text.afterLast = NULL
        destWalker->next = NULL
        destWalker->reserved = NULL

        if destUri->pathHead == NULL:
            destUri->pathHead = destWalker
            doneMask |= URI_NORMALIZE_PATH

        if copy range failed:
            prevent leak
            return ....

        if destPrev != NULL:
            destPrev->next = destWalker
        destPrev = destWalker
        sourceWalker = sourceWalker->next

    destUri->pathTail = destPrev
```

Some key differences/ideas are:
- single place with malloc
- single place with range copy
- dest walker is now clear to record the parent (and hence be one-off of source walker)
- `.first` and `.afterLast` set to NULL before trying to copy.

There is a chance there are bugs in this, since it's not implemented or tested yet, just a draft.

hartwork:
@kocsismate PS: I have gone forward merging #231 just now because…

- it was no longer in draft state and I felt sure enough of it by now
- it allows you to focus on the copy feature(but post-merge review is still welcome if you find time)
- it allows rebasing #230 here onto latest master with `URI_NORMALIZE_PORT` already in place
- it adds no public API other than enum value `URI_NORMALIZE_PORT` so making adjustments post merge has few constraints

I hope that's good with you?

include/uriparser/UriBase.h
```diff
@@ -322,7 +323,7 @@ URI_PUBLIC int uriCompleteMemoryManager(UriMemoryManager * memory,
  * @since 0.9.0
  */
 URI_PUBLIC void * uriEmulateCalloc(UriMemoryManager * memory,
-		size_t nmemb, size_t size);
+								   size_t nmemb, size_t size);
```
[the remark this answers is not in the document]
kocsismate: Hmm, not sure what went wrong again. github desktop hides whitespace changes for me by default. I removed this setting now, so I can notice these better...

src/UriCopy.c
```diff
@@ -0,0 +1,238 @@
...
+	}
+
+	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->portText, &sourceUri->portText, URI_FALSE, memory) == URI_FALSE) {
+		URI_FUNC(PreventLeakageAfterCopy)(destUri, doneMask, memory);
+		return URI_ERROR_MALLOC;
+	}
+
+	doneMask |= URI_NORMALIZE_PORT;
+
+	if (sourceUri->pathHead != NULL && sourceUri->pathTail != NULL) {
```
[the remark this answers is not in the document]
kocsismate: Yes, I wrote these additional conditions for pedanticality (maybe they once come handy for a static analyser), but they are not necessary indeed

kocsismate:
> I find this branch of the code overly complex even before being bullet proof

I agree, I needed quite some time to come up with it and understand it. So I'm happy with your suggestion. I'll try to implement it asap, but I'm on vacation, so my response time may be even more hectic than usuallly.

hartwork:
@kocsismate I'm looking forward to the next iteration. No need to hectify your vacation for me. I myself will also have some off-screen time in the coming 10 days. My worst expected response time will probably be 48 hours.

src/UriCopy.c
kocsismate: Works like a charm based on my tests!

kocsismate:
As far as I see, all the existing review comments are now addressed. I'm now working on the tests.

src/UriCopy.c
```diff
@@ -0,0 +1,234 @@
...
+			}
+
+			if (destPrev != NULL) {
+				destPrev->next = destWalker;
+			}
+
+			destPrev = destWalker;
+			sourceWalker = sourceWalker->next;
+
+			destUri->pathTail = destPrev;
```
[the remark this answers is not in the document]
kocsismate: Yes, I chose my original algorithm because it didn't have those steps inside the loop that are unnecessarily repeated, but it was very difficult to understand even for me. Now, I tried to move the creation of the head out of the loop by using the extract method approach, but it also resulted in a much more complicated code, and the result didn't seem to be very efficient performance-wise.

> PS: I was also considering to suggest to move the line outside of the loop but it's a tradeoff:

Since the `PreventLeakage` method doesn't depend on the `pathTail`, this optimization could still be implemented, couldn't it?

test/copy.cpp
```diff
@@ -0,0 +1,170 @@
...
+		state.uri = &sourceUri;
+
+		int result = uriParseSingleUriExA(&state, sourceUriString, sourceUriString + strlen(sourceUriString), NULL);
+		ASSERT_EQ(URI_SUCCESS, result);
+
+		UriUriA destUri;
+		ASSERT_EQ(URI_SUCCESS, uriCopyUriA(&destUri, &sourceUri));
+		ASSERT_EQ(URI_TRUE, uriEqualsUriA(&destUri, &sourceUri));
+
+		uriFreeUriMembersA(&sourceUri);
```
hartwork, line 47:
```suggestion
		uriFreeUriMembersA(&sourceUri);
		uriFreeUriMembersA(&destUri);
```
Or am I missing something?

test/copy.cpp
```diff
@@ -0,0 +1,170 @@
...
+	ASSERT_EQ(0, destUri->hostData->ipv6[6]);
+	ASSERT_EQ(0, destUri->hostData->ipv6[7]);
+	ASSERT_EQ(0, destUri->hostData->ipv6[8]);
+	ASSERT_EQ(0, destUri->hostData->ipv6[9]);
+	ASSERT_EQ(10, destUri->hostData->ipv6[10]);
+	ASSERT_EQ(185, destUri->hostData->ipv6[11]);
+	ASSERT_EQ(192, destUri->hostData->ipv6[12]);
+	ASSERT_EQ(168, destUri->hostData->ipv6[13]);
+	ASSERT_EQ(1, destUri->hostData->ipv6[14]);
+	ASSERT_EQ(2, destUri->hostData->ipv6[15]);
```
hartwork, line 152: @kocsismate based on typo "ipv6" instead of "ip6" I am now assuming that you only pushed for backup and not for review… please ping back when ready for review :pray:

kocsismate: Sorry, I wanted to try out my changes in CI, but I feel asleep immediately, and I didn't have time to fix my changes. You are right, I should have tried them out locally, because there were a lot of issues with it. Now, the tests succeed locally, but I'll have another look later, so please do not review it yet.

test/copy.cpp
kocsismate: Freeing happens on the call site, because the callers are asserting the components.

kocsismate:
@hartwork I should have fixed all your comments by now. I'll wait until you have the chance to review my latest commit, and then I'll reorder my commits by creating a separate commit for the `CopyRangeEngine() refactoring`.

test/copy.cpp
hartwork: @kocsismate the test is the call site — no? Please help me understand.

test/copy.cpp
```diff
@@ -114,42 +112,29 @@ TEST(CopyUriSuite, SuccessIpV4) {
 	UriUriA destUri;
 	testCopyUri(&destUri, "http://192.168.0.1/");
 
-	ASSERT_EQ(192, destUri.hostData.ip4->data[0]);
-	ASSERT_EQ(168, destUri.hostData.ip4->data[1]);
-	ASSERT_EQ(0, destUri.hostData.ip4->data[2]);
-	ASSERT_EQ(1, destUri.hostData.ip4->data[3]);
+	const unsigned char expected[4] = {192, 168, 0, 1};
+	ASSERT_EQ(memcmp(expected, destUri.hostData.ip4->data, 4 * sizeof(char)), 0);
```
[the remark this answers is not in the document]
kocsismate: I've just reordered my commits, and addressed these two comments.

test/copy.cpp
```diff
@@ -0,0 +1,153 @@
...
+	ASSERT_EQ(0, strncmp(destUri.hostText.first, "192.168.0.1", strlen("192.168.0.1")));
+
+	uriFreeUriMembersA(&destUri);
+}
+
+TEST(CopyUriSuite, SuccessIpV6) {
+	UriUriA destUri;
+	testCopyUri(&destUri, "https://[2001:0db8:0001:0000:0000:0ab9:c0a8:0102]");  // RFC 3849
+
+	const unsigned char expected[16] = {32, 1, 13, 184, 0, 1, 0, 0, 0, 0, 10, 185, 192, 168, 1, 2};
```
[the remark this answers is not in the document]
kocsismate: I implemented the suggestion by amending my 2nd commit. Nothing else was changed.

PR #234, opened 2025-06-23
Add function `HasHost` to the public API

kocsismate:
I need this function for prepending the `/` to the path component when the relevant getter is used (https://github.com/php/php-src/pull/18836/files#diff-04ea97756e1ce310ccfd4d8775089274da9a7ce11af774336dbda793e6642fcbR197):

```
if (uriparser_uri->absolutePath || uriIsHostSetA(uriparser_uri)) {
    smart_str_appendc(&str, '/');
}

...
```

### DIFF
--- a/include/uriparser/Uri.h
+++ b/include/uriparser/Uri.h
@@ -209,7 +209,7 @@ typedef struct URI_TYPE(QueryListStruct) {
  *
  * @since 0.9.9
  */
-URI_PUBLIC UriBool URI_FUNC(IsHostSet)(const URI_TYPE(Uri) * uri);
+URI_PUBLIC UriBool URI_FUNC(HasHost)(const URI_TYPE(Uri) * uri);
 
 
 
--- a/include/uriparser/Uri.h
+++ b/include/uriparser/Uri.h
@@ -201,6 +201,17 @@ typedef struct URI_TYPE(QueryListStruct) {
 } URI_TYPE(QueryList); /**< @copydoc UriQueryListStructA */
 
 
+/**
+ * Checks if a URI has the host component set.
+ *
+ * @param uri <b>IN</b>: %URI to check
+ * @return <c>URI_TRUE</c> when host is set, <c>URI_FALSE</c> otherwise
+ *
+ * @since 0.9.9
+ */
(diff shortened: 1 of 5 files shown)


include/uriparser/Uri.h
```diff
@@ -201,6 +201,15 @@ typedef struct URI_TYPE(QueryListStruct) {
 } URI_TYPE(QueryList); /**< @copydoc UriQueryListStructA */
 
 
+/**
+ * Checks if a URI has the host component set.
+ *
+ * @param uri %URI to check
```
hartwork, line 207: This line does not yet match the style of the rest of uriparser.

```suggestion
 * @param uri <b>IN</b>: %URI to check
```

include/uriparser/Uri.h
```diff
@@ -201,6 +201,15 @@ typedef struct URI_TYPE(QueryListStruct) {
...
+ * @return <c>URI_TRUE</c> when host is set, <c>URI_FAlSE</c> otherwise
```
hartwork, line 208:
```suggestion
 * @return <c>URI_TRUE</c> when host is set, <c>URI_FALSE</c> otherwise
 *
 * @since 0.9.9
```

include/uriparser/Uri.h
```diff
@@ -201,6 +201,15 @@ typedef struct URI_TYPE(QueryListStruct) {
...
+ */
+URI_PUBLIC UriBool URI_FUNC(IsHostSet)(const URI_TYPE(Uri) * uri);
```
hartwork, line 210: I'm considering a rename from `IsHostSet` to `HasHost`. What do you think?

hartwork (changes requested):

include/uriparser/Uri.h
kocsismate: I'm fine with whatever name you prefer, both make sense to me. Do you want me to  rename the function? Do you want two commits, or one is also fine (and then the commit message would be different)?

hartwork: @kocsismate two commits would be ideal: one for making public, one for the rename. Thanks!

hartwork (approved):

PR #237, opened 2025-07-01
Do not copy .hostData.ipFuture unnecessarily

kocsismate:
`.hostData.ipFuture` can hold the same range as `.hostText` does, and copying in this case is not necessary. Actually, PHP's tests revealed a memory leak without this fix when the copied and then normalized URI was used.

### DIFF
--- a/src/UriCopy.c
+++ b/src/UriCopy.c
@@ -153,7 +153,10 @@ int URI_FUNC(CopyUriMm)(URI_TYPE(Uri) * destUri,
 		*(destUri->hostData.ip6) = *(sourceUri->hostData.ip6);
 	}
 
-	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->hostData.ipFuture, &sourceUri->hostData.ipFuture, URI_FALSE, memory) == URI_FALSE) {
+	if (sourceUri->hostData.ipFuture.first != NULL && sourceUri->hostText.first == sourceUri->hostData.ipFuture.first) {
+		destUri->hostData.ipFuture.first = destUri->hostText.first;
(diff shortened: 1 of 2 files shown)


hartwork (commented):
@kocsismate thanks for bringing this up! I understand the the two hunks and the idea. I found related…

https://github.com/uriparser/uriparser/blob/5f7c6d88c50f548d0c7f499c22d36f51d34775b3/src/UriParse.c#L2289-L2294

…and…

https://github.com/uriparser/uriparser/blob/5f7c6d88c50f548d0c7f499c22d36f51d34775b3/src/UriNormalize.c#L137-L141

I believe you are saying that the latter would need to resemble former to not leak memory in `PreventLeakage` for URIs where `.hostText` and `.hostData.ipFuture` no longer point to the same range? If yes, that should probably be a dedicated pull request with a test cases (that fails without the fix and passes with the fix). Is that what you meant?

kocsismate:
> Is that what you meant?

Yes. When I was searching for the root cause of the memleak, I noticed that copying misses the optimization that's mentioned in the comment you linked in `UriParse.c`. After implementing this PR, I didn't have time to investigate the problem further, but I suspected it was going to be during normalization since that was the only relevant place I hadn't checked. I believe you found the exact place that needs to be fixed!

hartwork:
@kocsismate @TimWolla quick update on https://github.com/uriparser/uriparser/pull/237#issuecomment-3032358281 above and #241: I have started considering a 180° turn in the meantime and detect and reject mismatching `ipFuture` instead — still moving…

hartwork:
@kocsismate @TimWolla I have created pull request #243 just now for the going all-in with existing `.hostData.ipFuture` <-> `.hostText` relation constraints. Review welcome :pray:

PR #239, opened 2025-07-02
tests: Cover `FreeUriMembers` after `memset` with zeros (fixes #238)

### LINKED ISSUE #238: Is calling `FreeUriMembers` guaranteed to be safe for a zeroed `Uri`?
I'm reviewing @kocsismate’s integration of uriparser into PHP and the implementation embeds the `Uri` struct into a larger struct, where the embedded `Uri` is lazily initialized and thus does not necessarily hold any information. Currently we have a `bool` field to indicate whether or not the URI is initialized already:

```c
struct outer {
	UriUriA uri;
	UriUriA normalized_uri;
	bool normalized_uri_initialized;
};
```

I'm interested in simplifying the error paths and would like to be able to unconditionally `uriFreeUriMembersA(&outer.normalized_uri);` without checking whether the URI is initialized.

Thus I'd like to know if:

```c
UriUriA uri;
memset(&uri, 0, sizeof(uri));
uriFreeUriMembersA(&uri);
```

is guaranteed to safely not do anything? As far as I can tell from the implementation it is, but this guarantee is not spelled out in the doc comment for the `FreeUriMembers` function and thus might or might not change in the future.

hartwork:
Fixes #238

### DIFF
--- a/include/uriparser/Uri.h
+++ b/include/uriparser/Uri.h
@@ -334,6 +334,10 @@ URI_PUBLIC int URI_FUNC(ParseSingleUriExMm)(URI_TYPE(Uri) * uri,
  * itself is not freed, only its members.
  * Uses default libc-based memory manager.
  *
+ * @remarks
+ * Calling on an all-zeros structure (e.g. through <c>memset</c> or <c>calloc</c>) is safe.<br/>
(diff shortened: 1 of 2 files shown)


TimWolla (approved):
Thank you!

PR #240, opened 2025-07-03
copying: Add missing initial reset to destUri (follow-up to #230)

hartwork:
Follow-up to #230

Without this initial call, cleanup is shaky and e.g. a done mask with `URI_NORMALIZE_HOST` could cause function `PreventLeakage` to read uninitialized memory from `->hostData.ipFuture.first`.

CC @kocsismate @TimWolla


TimWolla (approved):
Makes sense to me as far as I understand it and is consistent with how parsing resets the target structure first.

PR #242, opened 2025-07-03
Percent encoding related cleanup

hartwork:
CC @kocsismate


hartwork:
@TimWolla @kocsismate thank you! :+1:

PR #243, opened 2025-07-06
Be consequent about and document `.hostData.ipFuture` `.hostText` relation

hartwork:


TimWolla (commented):
I don't feel confident actually “approving” this change, due to my lack of knowledge about the internals of this library. However the changes in this diff all make sense to me (I can't say whether you missed any, though).

PR #245, opened 2025-07-08
POC: Add support setting the scheme

kocsismate:
Just a small experiment... Not intended to be a full solution


src/UriSetter.c
```diff
@@ -0,0 +1,220 @@
...
+		UriMemoryManager * memory) {
+	if (uri == NULL) {
+		return URI_ERROR_NULL;
+	}
+
+	if (first != NULL && URI_FUNC(ValidateScheme)(first, afterLast) == URI_FALSE) {
+		return URI_ERROR_SYNTAX;
+	}
+
+	if (URI_FUNC(SetRange)(&uri->scheme, first, afterLast, uri->owner, memory) == URI_FALSE) {
```
hartwork, line 157: If this succeeds with `uri->owner` set but the rest of the function fails somewhere below, the duped scheme memory would be leaked.

src/UriSetter.c
```diff
@@ -0,0 +1,220 @@
...
+
+	URI_TYPE(Uri) *uri2 = memory->malloc(memory, sizeof(URI_TYPE(Uri)));
+	if (uriString == NULL) {
+		memory->free(memory, uriString);
+		return URI_ERROR_MALLOC;
+	}
+	int result = URI_FUNC(ParseSingleUriExMm)(uri2, uriString, uriString + bytesRequired - 1, NULL, memory);
+
+	URI_FUNC(FreeUriMembersMm)(uri2, memory);
+	memory->free(memory, uriString);
```
hartwork, line 187: Instead of re-parsing — please help me understand the motivation behind it better —, the function would likely need to:
- Disambiguate the path of the URI if (1) the first path segment contains a colon and (2) the scheme is about to be dropped.
- (Optionally/potentially dissolve disambiguation in path if a scheme is set and path starts with a dot segment followed by a segment containing a colon. Not fully sure yet).

PR #248, opened 2025-08-25
doc/Mainpage.txt: Add missing slash in example

gemmaro:
Hello,
This minor change adds a missing slash character to the example code block in the [Filenames and URIs](https://uriparser.github.io/doc/api/latest/index.html#filenames) section.
Thank you.

### DIFF
--- a/doc/Mainpage.txt
+++ b/doc/Mainpage.txt
@@ -154,7 +154,7 @@
  * 		free(absUri);
  * 		...
  * 	}
- * 	/COMMENT_HACK* absUri is "file:///E:/Documents%20and%20Settings" now *COMMENT_HACK/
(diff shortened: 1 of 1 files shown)


hartwork (changes requested):
@gemmaro the change suggested seems mistaken to me. I can explain why but I would rather you explain why you think the fix is needed if that's alright with you. Thanks!

gemmaro:
@hartwork Thank you for the explanation! You're absolutely right. I misread the filename example in the documentation - the two slashes are for escaping and actually represent a single slash.

Your explanation makes perfect sense - the current documentation example is correct as written. Thank you for taking the time to explain this clearly.

I'll close this PR now. Sorry for the confusion!

PR #249, opened 2025-08-25
Implement and test setter functions (fixes #196)

### LINKED ISSUE #196: [Feature request] Modification of URIs
I'm currently working on an RFC for the PHP language to use uriparser for parsing and manipulating URLs (https://wiki.php.net/rfc/url_parsing_api), and I realized during implementation that uriparser doesn't support modification of URL components once they are parsed.

As this feature could be potentially very useful for many users of the library, I figured that I draw your attention to this missing feature.

For comparison, ADA URL supports the same feature: https://cpp-api.ada-url.com/#autotoc_md18

hartwork:
Fixes #196

CC @kocsismate @pronskiy

### DIFF
--- a/include/uriparser/Uri.h
+++ b/include/uriparser/Uri.h
@@ -1549,6 +1549,10 @@ URI_PUBLIC UriBool URI_FUNC(IsWellFormedUserInfo)(const URI_CHAR * first, const 
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -1583,6 +1587,10 @@ URI_PUBLIC int URI_FUNC(SetFragment)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
@@ -1616,6 +1624,10 @@ URI_PUBLIC int URI_FUNC(SetFragmentMm)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * Uses default libc-based memory manager.
  *
@@ -1656,6 +1668,10 @@ URI_PUBLIC int URI_FUNC(SetHostAuto)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
@@ -1695,6 +1711,10 @@ URI_PUBLIC int URI_FUNC(SetHostAutoMm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -1728,6 +1748,10 @@ URI_PUBLIC int URI_FUNC(SetHostRegName)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
@@ -1763,6 +1787,10 @@ URI_PUBLIC int URI_FUNC(SetHostRegNameMm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -1796,6 +1824,10 @@ URI_PUBLIC int URI_FUNC(SetHostIp4)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
@@ -1831,6 +1863,10 @@ URI_PUBLIC int URI_FUNC(SetHostIp4Mm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -1864,6 +1900,10 @@ URI_PUBLIC int URI_FUNC(SetHostIp6)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
@@ -1899,6 +1939,10 @@ URI_PUBLIC int URI_FUNC(SetHostIp6Mm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -1932,6 +1976,10 @@ URI_PUBLIC int URI_FUNC(SetHostIpFuture)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
@@ -1968,6 +2016,10 @@ URI_PUBLIC int URI_FUNC(SetHostIpFutureMm)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * Uses default libc-based memory manager.
  *
@@ -2008,6 +2060,10 @@ URI_PUBLIC int URI_FUNC(SetPath)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
@@ -2045,6 +2101,10 @@ URI_PUBLIC int URI_FUNC(SetPathMm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -2081,6 +2141,10 @@ URI_PUBLIC int URI_FUNC(SetPortText)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
@@ -2114,6 +2178,10 @@ URI_PUBLIC int URI_FUNC(SetPortTextMm)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * Uses default libc-based memory manager.
  *
@@ -2149,6 +2217,10 @@ URI_PUBLIC int URI_FUNC(SetQuery)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
@@ -2183,6 +2255,10 @@ URI_PUBLIC int URI_FUNC(SetQueryMm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -2216,6 +2292,10 @@ URI_PUBLIC int URI_FUNC(SetScheme)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
@@ -2253,6 +2333,10 @@ URI_PUBLIC int URI_FUNC(SetSchemeMm)(URI_TYPE(Uri) * uri,
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
  *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
+ *
  * Uses default libc-based memory manager.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
@@ -2288,6 +2372,10 @@ URI_PUBLIC int URI_FUNC(SetUserInfo)(URI_TYPE(Uri) * uri,
  * or non-<c>NULL</c> at the same time.
  *
  * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * For all return values but <c>URI_ERROR_MALLOC</c>, all-or-nothing behavior
+ * can be expected, e.g. trying to apply a malformed value will leave the
+ * %URI unchanged.
  *
  * @param uri        <b>INOUT</b>: %URI to modify
  * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
--- a/include/uriparser/Uri.h
+++ b/include/uriparser/Uri.h
@@ -1197,6 +1197,7 @@ URI_PUBLIC int URI_FUNC(MakeOwnerMm)(URI_TYPE(Uri) * uri,
  * @param afterLast  <b>IN</b>: Pointer to character after the last one still in
  * @return           <c>URI_TRUE</c> if non-<c>NULL</c> and well-formed, else <c>URI_FALSE</c>
  *
+ * @see uriIsWellFormedHostIp4A
  * @see uriIsWellFormedHostRegNameA
  * @see uriIsWellFormedPortA
  * @see uriIsWellFormedQueryA
@@ -1210,6 +1211,25 @@ URI_PUBLIC UriBool URI_FUNC(IsWellFormedFragment)(const URI_CHAR * first, const 
 
 
 /**
+ * Determines if the given text range contains a well-formed IPv4 address
+ * according to RFC 3986 or not.
+ *
+ * @param first      <b>IN</b>: Pointer to first character
+ * @param afterLast  <b>IN</b>: Pointer to character after the last one still in
+ * @return           <c>URI_TRUE</c> if non-<c>NULL</c> and well-formed, else <c>URI_FALSE</c>
+ *
+ * @see uriIsWellFormedFragmentA
+ * @see uriIsWellFormedHostRegNameA
+ * @see uriIsWellFormedPortA
+ * @see uriIsWellFormedQueryA
+ * @see uriIsWellFormedUserInfoA
+ * @since 0.9.9
+ */
+URI_PUBLIC UriBool URI_FUNC(IsWellFormedHostIp4)(const URI_CHAR * first, const URI_CHAR * afterLast);
+
+
+
+/**
  * Determines if the given text range contains a well-formed registered host name
  * according to RFC 3986 or not.
  *
@@ -1218,6 +1238,7 @@ URI_PUBLIC UriBool URI_FUNC(IsWellFormedFragment)(const URI_CHAR * first, const 
  * @return           <c>URI_TRUE</c> if non-<c>NULL</c> and well-formed, else <c>URI_FALSE</c>
  *
  * @see uriIsWellFormedFragmentA
+ * @see uriIsWellFormedHostIp4A
  * @see uriIsWellFormedPortA
  * @see uriIsWellFormedQueryA
  * @see uriIsWellFormedUserInfoA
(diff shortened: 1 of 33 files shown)


kocsismate:
Thank you! I'm going to start the actual integration tomorrow :)

hartwork:
@kocsismate cool! :+1:

src/UriSetPort.c
```diff
@@ -0,0 +1,179 @@
...
+
+	/* Apply new value */
+	{
+		URI_TYPE(TextRange) sourceRange;
+		sourceRange.first = first;
+		sourceRange.afterLast = afterLast;
+
+		if (URI_FUNC(CopyRangeAsNeeded)(&uri->portText, &sourceRange, /*useSafe=*/ URI_TRUE, memory) == URI_FALSE) {
+			return URI_ERROR_MALLOC;
+		}
```
TimWolla, line 163: Okay, unfortunately in this case the implementation seems to leave the URI in a safe (i.e. it can safely be freed), but in a partly-modified state that was not requested.

From what I see it should be possible to keep the old `TextRange` and to only free it when the reassignment succeeded - and to restore it onto the object when the assignment of the new one fails.

----------------

The same remark appears to apply to the query and possibly others.

TimWolla: Something like this appears to do the trick:

```
	if ((first != NULL) && (URI_FUNC(IsWellFormedPort)(first, afterLast) == URI_FALSE)) {
		return URI_ERROR_SYNTAX;
	}

	/* Ensure owned */
	if (uri->owner == URI_FALSE) {
		const int res = URI_FUNC(MakeOwnerMm)(uri, memory);
		if (res != URI_SUCCESS) {
			return res;
		}
	}

	assert(uri->owner == URI_TRUE);

	/* Apply new value */
	{
		URI_TYPE(TextRange) destinationRange = uri->portText;

		URI_TYPE(TextRange) sourceRange;
		sourceRange.first = first;
		sourceRange.afterLast = afterLast;

		if (URI_FUNC(CopyRangeAsNeeded)(&uri->portText, &sourceRange, /*useSafe=*/ URI_TRUE, memory) == URI_FALSE) {
			return URI_ERROR_MALLOC;
		}

		/* Clear old value */
		if ((destinationRange.first != destinationRange.afterLast)) {
			memory->free(memory, (URI_CHAR *)destinationRange.first);
		}
	}
```

include/uriparser/Uri.h
```diff
@@ -1189,6 +1231,992 @@ URI_PUBLIC int URI_FUNC(MakeOwnerMm)(URI_TYPE(Uri) * uri,
...
+ * or non-<c>NULL</c> at the same time.
+ *
+ * The function may make the %URI own its memory if needed (if it is not already owned).
+ *
+ * Uses default libc-based memory manager.
+ *
+ * @param uri        <b>INOUT</b>: %URI to modify
+ * @param first      <b>IN</b>: Pointer to first character, can be <c>NULL</c>
+ * @param afterLast  <b>IN</b>: Pointer to character after the last one still in, can be <c>NULL</c>
+ * @return           Error code or 0 on success
```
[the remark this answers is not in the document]
hartwork: CC @kocsismate :arrow_up:

src/UriSetPort.c
TimWolla: From my cursory glance it seems like a simple reordering would enable the stronger guarantee, but indeed it would need additional tests if this should actually be guaranteed.

Since this is not necessary for PHP, feel free to resolve this thread if you feel that making the functions safe(r) in case of OOM is out of scope of the project.

src/UriCommon.c
```diff
@@ -83,6 +87,30 @@ void URI_FUNC(ResetUri)(URI_TYPE(Uri) * uri) {
...
+int URI_FUNC(FreeUriPath)(URI_TYPE(Uri) * uri, UriMemoryManager * memory) {
+	assert(uri != NULL);
+	assert(memory != NULL);
+
+	if (uri->pathHead != NULL) {
+		URI_TYPE(PathSegment) * segWalk = uri->pathHead;
+		while (segWalk != NULL) {
+			URI_TYPE(PathSegment) * const next = segWalk->next;
+			if (uri->owner && (segWalk->text.first != NULL)
+					&& (segWalk->text.first < segWalk->text.afterLast)) {
```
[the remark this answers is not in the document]
hartwork: @TimWolla I did see the inconsistency but didn't touch it earlier beyond a move because I was aiming for a pure function extraction refactoring. I have streamlined it now in commit 776c5dc31a75660691eb07a71a3f2a627b2fd5f6.

src/UriSetPort.c
TimWolla: @hartwork As part of reviewing the changes, I've tried out my suggestion and have a working patch for scheme, userinfo, port, path, query, and fragment (only host missing so far) with tests that ensures that the URI will not be partially modified if it previously is owned, no matter when malloc fails:

https://gist.github.com/TimWolla/99398bbeb56ec291f20755194f46774e

Would you be interested in including this patch? I've tried to follow the code style as closely as possible and to make a minimal amount of changes and hope that the patch reflects that. In fact the resulting code is simpler than before, because `CopyRangeAsNeeded()` already handled the case of a NULL range.

hartwork: @TimWolla I will look at your patch later today but please note that _my time budget on this is negative by now_, and I will not be able to fix anything but bugs affecting actual requirements from here on because of that. I appreciate the dialog with you but reviewing patches easily takes more time than building things myself and if I have to reject something both you and I will not feel great. No more patches for now please, in case you planned to put more into the pipeline. I'm sure you'll understand. Thanks! :pray:

TimWolla (commented):
I have now reviewed everything, except for:

- src/UriSetHostCommon.c
- src/UriSetPath.c
- test/SetPath.cpp

TimWolla:
Note: This requires a rebase + fix after #251.

hartwork:
> Note: This requires a rebase + fix after #251.

@TimWolla yes, coming up…

src/UriSetScheme.c
```diff
@@ -0,0 +1,271 @@
...
+	}
+
+	/* Clear old value */
+	if ((uri->owner == URI_TRUE) && (uri->scheme.first != uri->scheme.afterLast)) {
+		memory->free(memory, (URI_CHAR *)uri->scheme.first);
+	}
+	uri->scheme.first = NULL;
+	uri->scheme.afterLast = NULL;
+
+	/* Already done? */
```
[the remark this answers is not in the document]
TimWolla: I'm not sure I'm 100% happy, but that's definitely an improvement. In any case, not a deal breaker. Feel free to mark this thread as resolved (I can't myself, likely because I'm not an organization member).

src/UriSetScheme.c
```diff
@@ -0,0 +1,271 @@
...
+UriBool URI_FUNC(IsWellFormedScheme)(const URI_CHAR * first, const URI_CHAR * afterLast) {
+	if ((first == NULL) || (afterLast == NULL)) {
+		return URI_FALSE;
+	}
+
+	/* The related part of the grammar in RFC 3986 reads:
+	 *
+	 *   scheme      = ALPHA *( ALPHA / DIGIT / "+" / "-" / "." )
+	 */
+	if (first + 1 > afterLast) {
```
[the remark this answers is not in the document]
hartwork: @TimWolla that's an interesting find, and I I'm afraid you are right. My current plan is to use `afterLast - first < 1` instead to be "more defensive" and more self-explaining than `first == afterLast`. What do you think?

This is on `master`:

```console
# git --no-pager grep '\sif.*first +.*afterLast'
src/UriParse.c:         if (first + 1 >= afterLast) {
src/UriParse.c:                                         if (first + 1 >= afterLast) {
src/UriParse.c:                                                 if (first + 1 >= afterLast) {
src/UriParse.c:         if (first + 1 >= afterLast) {
src/UriParse.c:                 if (first + 2 >= afterLast) {
```

Pull request upcoming…

PS: I wonder why UBSan is not going off on these, I doubt that all of these are uncovered by tests :thinking:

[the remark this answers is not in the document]
hartwork: @TimWolla for branch `setters`, I applied these related changes now: https://github.com/uriparser/uriparser/compare/1ce7590a35ab985aa25130ecfa25d3814640be20..52bbf7173ba34e8f4bc23f22a0b861207cfed0fd . The idea was to rebase onto `master` after merge of #252 to join these two disjoint sets of fixes.

src/UriSetHostIp6.c
```diff
@@ -0,0 +1,182 @@
...
+		URI_CHAR candidate[3 + URI_MAX_IP6_LEN + 1 + 1] = _UT("//[");
+		const size_t inputLenChars = (afterLast - first);
+
+		/* Detect overflow */
+		if (inputLenChars > URI_MAX_IP6_LEN) {
+			return URI_ERROR_SYNTAX;
+		}
+
+		memcpy(candidate + 3, first, inputLenChars * sizeof(URI_CHAR));
+		memcpy(candidate + 3 + inputLenChars, _UT("]"), 2 * sizeof(URI_CHAR));
```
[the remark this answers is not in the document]
hartwork: @TimWolla exactly. I have added related comments just now:
https://github.com/uriparser/uriparser/compare/52bbf7173ba34e8f4bc23f22a0b861207cfed0fd..cc87e7d02c3ca184ac42fb4979d2f093918cdab8

test/SetHostRegName.cpp
```diff
@@ -0,0 +1,326 @@
...
+
+	uriFreeUriMembersA(&uri);
+}
+
+TEST(SetHostRegName, NullValueAppliedDotInserted) {
+	UriUriA uri = parseWellFormedUri("scheme://host//path1/path2");
+
+	EXPECT_EQ(uriSetHostRegNameA(&uri, NULL, NULL), URI_SUCCESS);
+
+	assertUriEqual(&uri, "scheme:/.//path1/path2");  // i.e. not scheme://path1/path2
```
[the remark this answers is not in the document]
hartwork: @TimWolla I do not see how `scheme:.//path1/path2` would have been an option, but it seems that we agree that there is nothing to fix here.

TimWolla:
I will take another look when the other PRs are merged and this one is rebased (to avoid GitHub's change tracking from getting confused with the rebase).

src/UriSetHostCommon.c
```diff
@@ -0,0 +1,258 @@
...
+		/* Clear old value */
+		const UriBool hadHostBefore = URI_FUNC(HasHost)(uri);
+		if (uri->hostData.ipFuture.first != NULL) {
+			if ((uri->owner == URI_TRUE) && (uri->hostData.ipFuture.first != uri->hostData.ipFuture.afterLast)) {
+				memory->free(memory, (URI_CHAR *)uri->hostData.ipFuture.first);
+			}
+			uri->hostData.ipFuture.first = NULL;
+			uri->hostData.ipFuture.afterLast = NULL;
+			uri->hostText.first = NULL;
+			uri->hostText.afterLast = NULL;
```
TimWolla, line 152: Should the:
```
			/* NOTE: .hostData.ipFuture holds the very same range pointers
			 *       as .hostText; we must not free memory twice. */
```
comment from UriParse.c be replicated here?

src/UriSetHostCommon.c
```diff
@@ -0,0 +1,258 @@
...
+			case URI_HOST_TYPE_IP4:
+				{
+					uri->hostData.ip4 = memory->malloc(memory, sizeof(UriIp4));
+					if (uri->hostData.ip4 == NULL) {
+						return URI_ERROR_MALLOC;
+					}
+
+					{
+						const int res = URI_FUNC(ParseIpFourAddress)(uri->hostData.ip4->data, first, afterLast);
+						assert(res == URI_SUCCESS);  /* because checked for well-formedness earlier */
```
TimWolla, line 222: I'd like to note that this causes a compiler warning if `NDEBUG` is example, since `res` is unused. Please see:

https://github.com/php/php-src/pull/19636/commits/1a5d94d762eb1bc8ab2318d622707bac194d9e50

The workaround in that commit doesn't seem quite right, though. A `(void)res;` seems to work:


```suggestion
						const int res = URI_FUNC(ParseIpFourAddress)(uri->hostData.ip4->data, first, afterLast);
						(void)res; /* suppress unused variable warning if assertions are disabled */
						assert(res == URI_SUCCESS);  /* because checked for well-formedness earlier */
```

test/SetPath.cpp
```diff
@@ -0,0 +1,408 @@
...
+	const char * const first = "///";
+	const char * const afterLast = first + strlen(first);
+
+	EXPECT_EQ(uriSetPathA(&uri, first, afterLast), URI_SUCCESS);
+
+	assertUriEqual(&uri, "scheme:/.///");  // i.e. not scheme:///
+	EXPECT_EQ(uri.absolutePath, URI_TRUE);
+
+	uriFreeUriMembersA(&uri);
+}
```
TimWolla, line 313: Should we have an explicit test case verifying the behavior for `file:` URIs? This seems to be a real world use case that folks expect to work and from what I see we don't have any path test with an empty host yet.

```
	UriUriA uri = parseWellFormedUri("file:///tmp/example.txt");
	const char * const first = "/root/example.txt";
	const char * const afterLast = first + strlen(first);

	EXPECT_EQ(uriSetPathA(&uri, first, afterLast), URI_SUCCESS);

	assertUriEqual(&uri, "file:///root/example.txt");
	EXPECT_EQ(uri.absolutePath, URI_FALSE);

	uriFreeUriMembersA(&uri);
```

test/SetPath.cpp
```diff
@@ -0,0 +1,408 @@
...
+
+TEST(SetPath, NonNullValueAppliedEmptyWithoutHost) {
+	UriUriA uri = parseWellFormedUri("scheme:path");
+	const char * const empty = "";
+
+	EXPECT_EQ(uriSetPathA(&uri, empty, empty), URI_SUCCESS);
+
+	assertUriEqual(&uri, "scheme:");
+	EXPECT_TRUE(uri.pathHead == NULL);
+	EXPECT_TRUE(uri.pathTail == NULL);
```
TimWolla, line 332: Should that be `EXPECT_EQ`?

src/UriSetPath.c
```diff
@@ -0,0 +1,493 @@
...
+
+		originalHead->text.first = NULL;
+		originalHead->text.afterLast = NULL;
+		memory->free(memory, originalHead);
+	}
+}
+
+
+
+static void URI_FUNC(TransformEmptyFirstPathSegment)(URI_TYPE(Uri) * uri, UriMemoryManager * memory) {
```
TimWolla, line 288: From naming and the lack of comments it's not quite clear to me what this function does from a high-level PoV.

src/UriSetPath.c
```diff
@@ -0,0 +1,493 @@
...
+	/* Skip the leading slash from target URIs with a host (so that we can
+	 * transfer the path 1:1 further down) */
+	if (URI_FUNC(HasHost)(destUri) == URI_TRUE) {
+		/* NOTE: This is because SetPathMm called IsWellFormedPath earlier: */
+		assert((afterLast - first >= 1) && (first[0] == _UT('/')));
+		first++;
+	} else if (first == afterLast) {
+	    /* This avoids (1) all the expensive but unecessary work below
+	     * and also (2) mis-encoding as single empty path segment
+	     * that would need (detection and) repair further down oetherwise */
```
TimWolla, line 332:
```suggestion
	     * that would need (detection and) repair further down otherwise */
```

src/UriSetPath.c
```diff
@@ -0,0 +1,493 @@
...
+					if (res != URI_SUCCESS) {
+						memory->free(memory, candidate);
+						return res;
+					}
+
+					/* Nothing but path and host is supposed to be set by the parse, in particular not: */
+					assert(tempUri.query.first == NULL);
+					assert(tempUri.fragment.first == NULL);
+
+					/* Ensure that the strings in the path segments are all owned */
```
TimWolla, line 383: I stumbled upon this and asked myself whether this is necessary, since `candidate` is already allocated by the correct MM. But it is necessary, because `candidate` is one large string and `MakeOwnerMm` splits this into multiple small strings, right?

Consider spelling this out more explicitly if I understood it correctly.

TimWolla (commented):
Went through it all now. Found just some minor nits and questions for clarification. Did not see any (remaining) actual bugs, but did not cross-reference the RFC for everything.

Please note that I am not doing the integration into PHP (I'm just reviewing it) and I am reviewing the upstream implementation as part of that as a courtesy to @kocsismate. Máté will have the final say in whether or not this implementation matches his requirements.

src/UriSetHostIp6.c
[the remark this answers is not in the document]
TimWolla: This thread can be resolved.

src/UriSetScheme.c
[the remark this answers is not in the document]
TimWolla: This thread can be resolved.

src/UriSetHostCommon.c
TimWolla: > why now…

Only started looking at that file in the last review iteration due to complexity / available time.

src/UriSetHostCommon.c
```diff
@@ -0,0 +1,258 @@
...
+	 * So no user info or port without a host. */
+	if (first == NULL) {
+		if (uri->userInfo.first != NULL) {
+			return URI_ERROR_SETHOST_USERINFO_SET;
+		} else if (uri->portText.first != NULL) {
+			return URI_ERROR_SETHOST_PORT_SET;
+		}
+	}
+
+	if (first != NULL) {
```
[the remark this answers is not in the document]
TimWolla: No strong opinion from my side. That's why the comment included the question-mark.

src/UriSetHostCommon.c
hartwork: @TimWolla that's an interesting one! For one, because it raises the question whether the uriparser upstream CI matrix should cover the `NDEBUG` case. For two, it makes me wonder why PHP is compiling with fatal warnings when it is third party code. With compilers getting stricter by the minute (for good and for bad) that seems like asking for future trouble. What do you think? PS: I'll add a variant of the `(void)res;` with the next push…

test/SetPath.cpp
hartwork: @TimWolla the `file` scheme is by no means special to RFC 3986, so it did not get any dedicated test cases. I have still added a testcase `SetPath.NonNullValueAppliedNonEmptyWithEmptyHost` for you now that "happens" to use the `file` scheme so that you can be at ease about both the empty host and `file` URIs working as expected. Push upcoming…

src/UriSetHostCommon.c
TimWolla: > For two, it makes me wonder why PHP is compiling with fatal warnings when it is third party code. With compilers getting stricter by the minute (for good and for bad) that seems like asking for future trouble. What do you think?

The build system of PHP is a complex autoconf-based beast. I can't claim to understand it :smiley: Building everything with `-Werror` in CI makes sense to me, since (expected) warnings from vendored libraries make it harder to review the CI logs in case of legit warnings for PHP's own code. In case of legit warnings that are caught by PHP's CI, they should be fixed upstream and in other cases, PHP is likely going to carry a patch.

For reference, the relevant config.m4 is this one: https://github.com/php/php-src/blob/master/ext/uri/config.m4

test/SetPath.cpp
hartwork: @TimWolla if I remember correctly, `EXPECT_EQ` with null pointers needs `nullptr` but we do not have C++11 in this code base as of today. Here's what the change to `EXPECT_EQ` with `NULL` would give us:

```console
# CTEST_OUTPUT_ON_FAILURE=1 make -C build_fuzzers_asan_ubsan/ all test
make: Entering directory '[..]/build_fuzzers_asan_ubsan'
[ 25%] Built target uriparser
[ 27%] Built target uriparse
[ 28%] Building CXX object CMakeFiles/testrunner.dir/test/SetPath.cpp.o
In file included from [..]/test/SetPath.cpp:25:
/usr/include/gtest/gtest.h:1394:11: error: comparison between pointer and integer ('UriPathSegmentStructA *' and 'long')
 1394 |   if (lhs == rhs) {
      |       ~~~ ^  ~~~
/usr/include/gtest/gtest.h:1413:12: note: in instantiation of function template specialization 'testing::internal::CmpHelperEQ<UriPathSegmentStructA *, long>' requested here
 1413 |     return CmpHelperEQ(lhs_expression, rhs_expression, lhs, rhs);
      |            ^
[..]/test/SetPath.cpp:331:2: note: in instantiation of function template specialization 'testing::internal::EqHelper::Compare<UriPathSegmentStructA *, long, nullptr>' requested here
  331 |         EXPECT_EQ(uri.pathHead, NULL);
      |         ^
/usr/include/gtest/gtest.h:1885:54: note: expanded from macro 'EXPECT_EQ'
 1885 |   EXPECT_PRED_FORMAT2(::testing::internal::EqHelper::Compare, val1, val2)
      |                                                      ^
1 error generated.
make[2]: *** [CMakeFiles/testrunner.dir/build.make:219: CMakeFiles/testrunner.dir/test/SetPath.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:182: CMakeFiles/testrunner.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
make: Leaving directory '[..]/build_fuzzers_asan_ubsan'

```

test/SetPath.cpp
TimWolla: > the `file` scheme is by no means special to RFC 3986

Yes, I understand. This was perhaps badly phrased from my side, the `file` scheme was just a convenient way to showcase this particular case of three consecutive slashes / empty hostnames. Thanks!

src/UriSetPath.c
hartwork: @TimWolla good catch. I should add codespell to CI here some time also, happily using that elsewhere.

test/SetPath.cpp
TimWolla: It seems you are right. I thought that worked, but perhaps I was confusing that with the output for an unexpected NULL when comparing two pointers.

src/UriSetPath.c
hartwork: @TimWolla improved comment coming up with next push for re-review…

src/UriSetHostAuto.c
```diff
@@ -0,0 +1,152 @@
...
+							return res;
+						}
+						hostType = URI_HOST_TYPE_IP6;
+					}
+					break;
+			}
+		/* IPv4? (else: RegName) */
+		} else if (URI_FUNC(IsWellFormedHostIp4)(first, afterLast)) {
+			hostType = URI_HOST_TYPE_IP4;
+		}
```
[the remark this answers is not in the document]
TimWolla: This thread is resolved.

test/SetPath.cpp
hartwork: @TimWolla I'll mark this as resolved already, please unresolve or comment again if the just-pushed test case has any issues.

test/SetPath.cpp
TimWolla: This thread is resolved.

src/UriSetHostCommon.c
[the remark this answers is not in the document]
TimWolla: This thread is resolved.

src/UriSetHostCommon.c
TimWolla: This thread is resolved (unless you have follow-up questions).

hartwork: @TimWolla I'll resist going too much more into that topic to not derail the review, but would like to be clear that for (both libexpat and) uriparser I am happy to address warnings upstream *but* I'll want bug reports rather than patches for compiler warnings in the future please, and address them myself.

src/UriSetPath.c
```diff
@@ -0,0 +1,499 @@
...
+
+					/* Disambiguate as needed */
+					{
+						const UriBool success = URI_FUNC(FixPathNoScheme)(destUri, memory);
+						if (success == URI_FALSE) {
+							return URI_ERROR_MALLOC;
+						}
+					}
+					{
+					    const UriBool success = URI_FUNC(EnsureThatPathIsNotMistakenForHost)(destUri, memory);
```
TimWolla, line 425: Indentation mismatch.

src/UriSetPath.c
```diff
@@ -0,0 +1,499 @@
...
+
+		originalHead->text.first = NULL;
+		originalHead->text.afterLast = NULL;
+		memory->free(memory, originalHead);
+	}
+}
+
+
+
+/* URIs with out a host encode a leading slash in the path as .absolutePath == URI_TRUE.
```
TimWolla, line 288:
```suggestion
/* URIs without a host encode a leading slash in the path as .absolutePath == URI_TRUE.
```

src/UriSetPath.c
TimWolla: I'm afraid, the purpose of the second call to `DropEmptyFirstPathSegment` is still not quite clear to me (why does it need to drop two empty segments?). I can see that test cases fail without it, so it must do something important, but it's not clear from following the logic why.

Is this related to how the path is parsed by prepending `///` (i.e. empty host + leading slash)?

src/UriSetPath.c
TimWolla: This thread is resolved.

src/UriSetPath.c
TimWolla: This thread is resolved.

include/uriparser/Uri.h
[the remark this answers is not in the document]
TimWolla: Greatly appreciated, looking good to me. Thank you.

This thread is resolved.

TimWolla (approved):
GitHub's new PR review interface is not showing me the list of pending comments when working with multiple tabs. Hopefully it did not eat anything.

I've found a few final nits and one case that is still not clear to me (but that might be on me). The implementation is good to me. Thank you for bearing with me 🙏

TimWolla:
> GitHub's new PR review interface is not showing me the list of pending comments when working with multiple tabs. Hopefully it did not eat anything.

Thankfully the list looks complete.

src/UriSetHostIpFuture.c
```diff
@@ -0,0 +1,174 @@
...
+				{
+					URI_TYPE(Uri) uri;
+					const int res = URI_FUNC(ParseSingleUriExMm)(&uri, candidate, candidate + candidateLenChars, NULL, memory);
+
+					assert((res == URI_SUCCESS) || (res == URI_ERROR_SYNTAX) || (res == URI_ERROR_MALLOC));
+
+					if (res == URI_SUCCESS) {
+						assert(uri.hostData.ipFuture.first != NULL);
+						URI_FUNC(FreeUriMembersMm)(&uri, memory);
+					}
```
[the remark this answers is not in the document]
TimWolla: Since you used “tending” (which leaves room for uncertainty), I'll reply: I quite like being able to call destructors unconditionally (just like `free()` guarantees to be safe with `NULL`), since this makes the code much less verbose. That's also why I filed that issue back then.

But of course you decide on the code style within the library you maintain :smiley:

test/SetHostRegName.cpp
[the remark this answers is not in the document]
TimWolla: > but it seems that we agree that there is nothing to fix here.

Yes. Except for a possible comment as to why you don't see the other output as an option (is there a clear reference in the RFC?). I'm afraid I'm unable to explain my train of thought, though.

hartwork:
@TimWolla I'll get back to your latest comments around 01:00 UTC+2…

TimWolla:
> I'll get back to your latest comments around 01:00 UTC+2…

Same timezone for me (Münsterland area), I'll likely be in bed by then :smile:

src/UriSetHostIpFuture.c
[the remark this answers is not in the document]
hartwork: @TimWolla I'm with you in general, but I misdesigned this part ~18 years ago, I cannot break API and the original design makes it dangerous to advise users to call FreeUriMembers in all failure cases by themselves (without checking which) and so I rather stick with dogfooding that decision until I start work on uriparser v3 if that happens. (Name uriparser2 is used by others so I'll skip v2.)

src/UriSetPath.c
hartwork: @TimWolla good catch, fixed with next push…

include/uriparser/Uri.h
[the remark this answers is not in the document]
hartwork: @TimWolla excellent! :tada:

src/UriSetPath.c
TimWolla: This thread is resolved.

test/SetPath.cpp
```diff
@@ -0,0 +1,421 @@
...
+	int charsRequired = -1;
+	ASSERT_EQ(uriToStringCharsRequiredA(uri, &charsRequired), URI_SUCCESS);
+	ASSERT_TRUE(charsRequired >= 0);
+
+	char * const buffer = (char *)malloc(charsRequired + 1);
+	ASSERT_TRUE(buffer != NULL);
+
+	ASSERT_EQ(uriToStringA(buffer, uri, charsRequired + 1, NULL), URI_SUCCESS);
+
+	ASSERT_EQ(strcmp(buffer, expected), 0);
```
TimWolla, line 58: I've tried around with the test suite a little more to fully understand `TransformEmptyLeadPathSegments` and the correctness of `SetPath()` in general. While doing so, I learned that this works, and returns better error messages:

```suggestion
	ASSERT_STREQ(buffer, expected);
```

results in:
```
Expected equality of these values:
  buffer
    Which is: "scheme:/./"
  expected
    Which is: "scheme:/"
```
when deleting the removal of the second empty segment.

test/SetHostRegName.cpp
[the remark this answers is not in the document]
TimWolla: > our input is an absolute path

I think the trouble I'm having is with this bit. My understanding is that the distinction between an absolute path and a relative path does not matter once the URI has a hostname, so from my mental model, the path in this case is neither absolute, nor relative.

But given that we both agree that the implementation is fine, I'm happy to leave it at.

src/UriSetPath.c
TimWolla: Okay, this specifically affects the `/` input only (i.e. setting an empty absolute path) and is necessary because otherwise the call to `EnsureThatPathIsNotMistakenForHost` would transform that into `/./` (which to my understanding would not technically be wrong, but it would clearly violate user expectations).

I think the troubles I was having with following the logic is that this is “working around” the problem at the wrong location: `EnsureThatPathIsNotMistakenForHost` is being called despite it being impossible for the path to be mistaken for a host.

An alternative implementation that passes all tests and that I find clearer (since there is a clear split of responsibilities) would be the following:

```diff
diff --git i/src/UriCommon.c w/src/UriCommon.c
index cf9530e..a635b76 100644
--- i/src/UriCommon.c
+++ w/src/UriCommon.c
@@ -767,6 +767,7 @@ UriBool URI_FUNC(EnsureThatPathIsNotMistakenForHost)(URI_TYPE(Uri) * uri,
        if ((URI_FUNC(HasHost)(uri) == URI_TRUE)
                        || (uri->absolutePath == URI_FALSE)
                        || (uri->pathHead == NULL)
+                       || (uri->pathHead == uri->pathTail)
                        || (uri->pathHead->text.first != uri->pathHead->text.afterLast)) {
                return URI_TRUE;  /* i.e. nothing to do */
        }
diff --git i/src/UriSetPath.c w/src/UriSetPath.c
index 78e7177..0d19c1a 100644
--- i/src/UriSetPath.c
+++ w/src/UriSetPath.c
@@ -307,10 +307,6 @@ static void URI_FUNC(TransformEmptyLeadPathSegments)(URI_TYPE(Uri) * uri, UriMem
        URI_FUNC(DropEmptyFirstPathSegment)(uri, memory);
 
        uri->absolutePath = URI_TRUE;
-
-       if ((uri->pathHead == uri->pathTail) && (uri->pathHead->text.first == uri->pathHead->text.afterLast)) {
-               URI_FUNC(DropEmptyFirstPathSegment)(uri, memory);
-       }
 }
```

I'm leaving it up to you to decide whether or not to take my suggested change. I can confirm that I now understand the logic.

hartwork: @TimWolla I struggled to understand your idea at first… but now got it. Very nice! :beers: Upcoming with the next push…

src/UriSetPath.c
```diff
@@ -0,0 +1,502 @@
...
+					/* Restore use of .absolutePath and fix //// as needed */
+					URI_FUNC(TransformEmptyLeadPathSegments)(destUri, memory);
+
+					/* Disambiguate as needed */
+					{
+						const UriBool success = URI_FUNC(FixPathNoScheme)(destUri, memory);
+						if (success == URI_FALSE) {
+							return URI_ERROR_MALLOC;
+						}
+					}
```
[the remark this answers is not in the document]
hartwork: @TimWolla understood and fixed, kudos for this finding :+1:

TimWolla (approved):
I found one last typo in a comment. Other than that, I don't have any further comments. With the latest changes, I understand the entire logic and I am not seeing any further bugs.

hartwork:
> I found one last typo in a comment. Other than that, I don't have any further comments. With the latest changes, I understand the entire logic and I am not seeing any further bugs.

@TimWolla cool! Thanks for being friendly with me and kudos in particular for the real bugs you uncovered! :+1: :pray:

src/UriSetPort.c
```diff
@@ -0,0 +1,179 @@
...
+	if ((uri == NULL) || ((first == NULL) != (afterLast == NULL))) {
+		return URI_ERROR_NULL;
+	}
+
+	URI_CHECK_MEMORY_MANAGER(memory);  /* may return */
+
+	/* The RFC 3986 grammar reads:
+	 *   authority = [ userinfo "@" ] host [ ":" port ]
+	 * So no port without a host. */
+	if (URI_FUNC(HasHost)(uri) == URI_FALSE) {
```
kocsismate, line 123: I think this should allow the port to be cleared. In PHP, this is the following code:

```
$uri = Uri\Rfc3986\Uri::parse("/foo");
$uri1->withPort(null); // TODO should not throw
```

kocsismate:
Integration is basically ready at https://github.com/php/php-src/pull/19636. Of course some things still need to be fixed, but otherwise it's fairy complete.

src/UriSetHostCommon.c
```diff
@@ -0,0 +1,266 @@
...
+		} else if (uri->portText.first != NULL) {
+			return URI_ERROR_SETHOST_PORT_SET;
+		}
+	}
+
+	/* Syntax-check the new value */
+	if (first != NULL) {
+		switch (hostType) {
+			case URI_HOST_TYPE_IP4:
+				if (URI_FUNC(IsWellFormedHostIp4)(first, afterLast) == URI_FALSE) {
```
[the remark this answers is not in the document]
hartwork: @kocsismate pushed.

hartwork:
@kocsismate FYI I'm currently planning on releasing 0.9.9 off of `master` with setters September 3rd — Wednesday — in the afternoon of UTC+2.

PR #250, opened 2025-08-30
src/UriCopy.c: Fix copying URIs with empty components

TimWolla:
When trying to copy a URI with empty userInfo, portText, query, or fragment components, the copying process would perform a zero-byte allocation for the respective component.

Depending on the malloc implementation, this would either return `NULL`, leading to a bogus `URI_ERROR_MALLOC` being returned from the copying process or a valid zero-byte allocation that will not be freed by FreeUriMembers, resulting in a memory leak.

### DIFF
--- a/src/UriCopy.c
+++ b/src/UriCopy.c
@@ -119,7 +119,7 @@ int URI_FUNC(CopyUriMm)(URI_TYPE(Uri) * destUri,
 
 	revertMask |= URI_NORMALIZE_SCHEME;
 
-	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->userInfo, &sourceUri->userInfo, URI_FALSE, memory) == URI_FALSE) {
+	if (URI_FUNC(CopyRangeAsNeeded)(&destUri->userInfo, &sourceUri->userInfo, URI_TRUE, memory) == URI_FALSE) {
 		URI_FUNC(PreventLeakageAfterCopy)(destUri, revertMask, memory);
 		return URI_ERROR_MALLOC;
 	}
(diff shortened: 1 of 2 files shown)


hartwork (commented):
@TimWolla I'm confirming the leak, thank you for the report and the patch candidate! :+1: :+1: 

This change seems to (rightfully) render the `useSafe` flag parameter of function `CopyRangeAsNeeded` pointless (despite the few places still passing `URI_FALSE`). A follow-up commit to resolve that flag (to effective `URI_TRUE` everywhere) would be cool… but I can also do that myself as a follow-up pull request, as you prefer. What do you think?

TimWolla:
> A follow-up commit to resolve that flag (to effective `URI_TRUE` everywhere) would be cool… but I can also do that myself as a follow-up pull request, as you prefer. What do you think?

The CopyRangeAsNeeded function is not yet within any release version (it does not exist in 0.9.8). Should I remove the flag entirely or should I ignore it? I'm happy to do either in a follow-up PR.

hartwork:
> > A follow-up commit to resolve that flag (to effective `URI_TRUE` everywhere) would be cool… but I can also do that myself as a follow-up pull request, as you prefer. What do you think?
> 
> The CopyRangeAsNeeded function is not yet within any release version (it does not exist in 0.9.8). Should I remove the flag entirely or should I ignore it? I'm happy to do either in a follow-up PR.

@TimWolla I was thinking of fully dissolving the parameter as if it was always `URI_TRUE`. The function is not public API so we cannot break API or ABI there. Thank you!

CC @kocsismate

TimWolla:
> The function is not public API so we cannot break API or ABI there. Thank you!

I see, the header is in src/, not in include/. Will send the follow-up once this one is merged.

hartwork:
> > The function is not public API so we cannot break API or ABI there. Thank you!
> 
> I see, the header is in src/, not in include/. Will send the follow-up once this one is merged.

@TimWolla deal :handshake:

hartwork (approved):

PR #251, opened 2025-08-30
src/UriCommon.h: Remove `CopyRangeAsNeeded`'s `useSafe` parameter

TimWolla:
Since the previous commit 00fd627db26475981fa7470e599c325a0934b900 that fixes a memory leak when copying URIs with empty components, the `useSafe` parameter of the `CopyRangeAsNeeded` function is always either set to `URI_TRUE` or the function is called with a non-empty range, which bypasses the `useSafe` parameter entirely.

Since this parameter is useless, we can remove it entirely, removing a possible footgun from the API.


hartwork (approved):
@TimWolla sweet, thank you! :pray: :beers: 

CC @kocsismate

TimWolla:
You're welcome. Thanks for the quick turn-around times and for maintaining this library :smiley:

PR #252, opened 2025-08-31
`src/UriParse.c`: Dissolve undefined behavior

hartwork:
In reaction to https://github.com/uriparser/uriparser/pull/249#discussion_r2312063320 …

Pointer arithmetic beyond the end (one after the last) of the array even without dereferencing is undefined behavior.

Affected functions are:
- `ParseIpFuture`
- `ParseIPv6address2`
- `ParsePctEncoded`

CC @TimWolla

### DIFF
--- a/src/UriParse.c
+++ b/src/UriParse.c
@@ -480,7 +480,7 @@ static const URI_CHAR * URI_FUNC(ParseIpFuture)(URI_TYPE(ParserState) * state,
 	case _UT('v'):
 	case _UT('V'):
 	*/
-		if (first + 1 >= afterLast) {
+		if (afterLast - first < 2) {
 			URI_FUNC(StopSyntax)(state, afterLast, memory);
 			return NULL;
 		}
@@ -755,7 +755,7 @@ static const URI_CHAR * URI_FUNC(ParseIPv6address2)(
 						}
 
 						/* "::"? */
-						if (first + 1 >= afterLast) {
+						if (afterLast - first < 2) {
 							URI_FUNC(StopSyntax)(state, afterLast, memory);
 							return NULL;
 						}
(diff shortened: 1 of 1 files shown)


src/UriParse.c
```diff
@@ -480,7 +480,7 @@ static const URI_CHAR * URI_FUNC(ParseIpFuture)(URI_TYPE(ParserState) * state,
 	case _UT('v'):
 	case _UT('V'):
 	*/
-		if (first + 1 >= afterLast) {
+		if (afterLast - first < 2) {
```
TimWolla, line 483: Should the `first >= afterLast` in line 470 be adjusted to `afterLast - first < 1` for uniform conditions?

TimWolla (approved):
The changes appear correct to me and from what I see you did not miss any locations.

src/UriParse.c
hartwork: @TimWolla good question. I added four cases of `afterLast - first < 1` on branch `setters` recently and I see 37 cases of `first >= afterLast` on `master`. Are you in clear favor of either approach? I'm tending towards using `first >= afterLast` everywhere (i.e. adjusting the `setters` branch once more) but the only key reason is "fewer changes". I'd even be okay with this not being consistent to the last bit. What do you think?

TimWolla: I don't have a strong opinion either way (that's why I selected “Approve”). I just wanted to make sure to diligently review the changes and note everything that came to mind. “Making fewer changes” is a reasonable choice.

hartwork: @TimWolla thanks for sharing your view! I will merge as-is here then and adjust #249 for consistency right after… :+1:

Issue #255, opened 2025-08-31
Release uriparser 0.9.9


hartwork:
Pull requests that motivate cutting this new release now:

- #249 

Other yet-to-merge pull requests going into the release:

- #252
- #253
- #254

Regular releases:

* [x] Changes in code:
  * [x] Bump soname
  * [x] Make sure that all change log items point to GitHub issues and/or pull requests
  * [x] Bump version everywhere
  * [x] Fill release date
  * [x] Update web-wide unique release identifier
* [x] On SourceForge:
  * [x] Upload files
  * [x] Mark new uploads as default downloads (`.zip` for Windows, `.tar.xz` all else)
* [x] On GitHub:
  * [x] Add and push Git tag
  * [x] Make a new release, upload files
  * [x] Associate GitHub issues and pull requests with the milestone of this release
  * [x] Close milestone on GitHub + create a new one
* [x] Update API docs at uriparser website — https://github.com/uriparser/uriparser.github.io/commit/1d603f13264f9e43d6cb76965b781c5ac8b4988d
* [x] Announce release
  * [x] uriparser website — https://github.com/uriparser/uriparser.github.io/commit/224306ffa6cde6c954e58abf827a98d8751efaef
  * [ ] ~uriparser-users — TODO~
  * [ ] ~blog.hartwork.org — TODO~
    * [ ] ~Hacker news — TODO~
  * [x] direct mail to distro maintainers
* [x] Bump Gentoo ebuild — https://github.com/gentoo/gentoo/commit/efead535a9d1fd0339c64491b834a25bb8c2974e

Specific to 0.9.9:

- n/a

Related downstreams issues and pull requests:

- n/a

CC @kocsismate @TimWolla

Issue #219, opened 2025-03-20
Is there a way to get version information at runtime?


FromeXo:
Hi,
First of all, thanks for maintaining uriparser!

I'm new to C programming, so apologies if I missed something obvious, but I was looking for a way to get the uriparser version at runtime. I saw that version numbers are defined in the UriBase.h using macros like:
```C
#define URI_VER_MAJOR 0
#define URI_VER_MINOR 9
#define URI_VER_RELEASE 8
```
From my understanding, these are only available at compile time. I searched the codebase for something like a function to retrieve the version at runtime but didn't find anything.

Would it be possible (or does it already exist?) to have the version information exposed at runtime? This would be useful for applications dynamically linking against uriparser to check which version they are running with.

Again, sorry if I missed something, and thanks for your time!

hartwork:
Hello @FromeXo,

from a quick check: I don't think there is a runtime version getter like that. The source of truth would often be the version provided by the package manager though because distros backport security fixes without increasing version numbers and so the string returned by a runtime version getter often does not reflect the reality with regard to what bugs were fixed, in particular not security ones. Where no package manager is used, the runtime and build version are likely identical?

With that in mind, what service would a runtime version getter provide to you?

Best, Sebastian

FromeXo:
Hi @hartwork,

Thanks for the quick response and the explanation! That makes a lot of sense, especially regarding package managers backporting fixes without changing the version number.

The main use case I was thinking of is for applications that want to log or display the uriparser version they are using - especially in cases where the library is dynamically linked. This could help with debugging or support cases where someone needs to quickly check which version is running without having to trace it through a package manager.

I totally understand if this isn't something that fits well with uriparser, but I just wanted to clarify my reasoning. Thanks again for your time!

Best,
FromeXo

hartwork:
@FromeXo thanks for your reply! 👍

hartwork:
Reopening…

PR #256, opened 2025-08-31
[DRAFT] Prepare release 0.9.9 (part of #255)

hartwork:
Part of #255

CC @kocsismate @TimWolla


ChangeLog
```diff
@@ -2,6 +2,103 @@ NOTE: uriparser is looking for help with a few things:
...
+      New functions:
+        uriHasHost[AW]
+  * Added: CMake: Add alias "uriparser::uriparser" (GitHub #197)
+  * Added: Integrate fuzzers from google/oss-fuzz repository and
+      improve those fuzzers on top (GitHub #209, GitHub #211, GitHub #212,
+      GitHub #214)
+      Thanks to @tyler92 for two of the related pull requests!
+  * Added: Support setting individual components of a UriUri[AW] structure
+      (GitHub #196, GitHub #249)
+      Thanks to Máté Kocsis and to Tim Düsterhus!
```
pronskiy, line 36: Can you also add that The PHP Foundation commissioned the work?

hartwork:
@TimWolla are you aware that because of security backports in distros runtime version information is often not accurate and would need to be read as "that version minimum plus an unknown amount of security fixes"? I have rejected a previous request for adding runtime version at https://github.com/uriparser/uriparser/issues/219#issuecomment-2741227520 because of that. Simplified it cannot be be relied upon and then what is the value? With that said, does it change the picture for you with PHP?

TimWolla:
> What if we have a function `uriMinimumRuntimeVersion` that returned string `0.9.9+` or `>=0.9.9` to be crystal clear that it's nothing but a minimum?

That would likely work for PHP's use case. To showcase how it will look like:

image

and on the CLI: 

```
uri

uri support => active
uriparser library version => 0.9.8
```

This is what we currently have on the `phpinfo();` output. I would probably split this into “compile time version” and “runtime version” to show both of them:

```
uri

uri support => active
Compiled against uriparser => 0.9.8
Running with uriparser => 0.9.9+
```

Distros would still be able to appropriately patch `uriMinimumRuntimeVersion()` to provide additional information if the expected guarantee by the function is “a string for human consumption”.

hartwork:
@TimWolla I became aware that the "+" suffix assumed additions when in fact distributors could also _cut away_ functionality where it would be misleading. So I discarded the addition of "+". Pull request #258 is ready for review, if you have a minute :pray:

TimWolla:
Thank you!

kocsismate:
Thanks!

hartwork:
Glad I could be of help! :pray:

PR #257, opened 2025-09-01
tests: Leverage assertion `EXPECT_STREQ`

hartwork:
In reaction to https://github.com/uriparser/uriparser/pull/249#discussion_r2314433948 …

Docs: https://google.github.io/googletest/reference/assertions.html#EXPECT_STREQ

CC @TimWolla


hartwork:
@TimWolla true. Ideally the pattern would be to only use `ASSERT_*` when the code has stop stop processing at the point and use `EXPECT_*`. The thing is that while tests pass the difference is tiny and making sure to pick the right one arguably costs more than just going with `ASSERT_*` everywhere. That and the code coming from another test frame that did not make the distinction…

PR #259, opened 2025-09-07
Codespell fixes and CI

hartwork:
CC @TimWolla


TimWolla (approved):
I'm not particularly familiar with Codespell, but all the spelling changes you made, make sense to me.

Issue #275, opened 2025-11-09
Removing host turns empty path into absolute path


TimWolla:
Originally reported at php/php-src#20431.

When removing the host from `https://example.com`, the result is `https:/`, but I expected `https:`.

In the original URI, the path matches the `path-abempty` production, which is “begins with "/" or is empty”. Since the path clearly doesn't begin with `/`, it must be empty. I am therefore expecting the path to still be empty after removing the host.

```
    UriUriA uri = parseWellFormedUri("https://example.com");

    EXPECT_EQ(uriSetHostAutoA(&uri, NULL, NULL), URI_SUCCESS);

    assertUriEqual(&uri, "https:");

    uriFreeUriMembersA(&uri);
```

hartwork:
@TimWolla I haven't reproduced it yet but I can follow your reasoning why you would want and expect `https:` for a result 👍

PR #276, opened 2025-11-09
src/UriSetHostCommon.c: Do not set `absolutePath` for empty paths when removing host

TimWolla:
Fixes uriparser/uriparser#275.

### DIFF
--- a/test/SetHostAuto.cpp
+++ b/test/SetHostAuto.cpp
@@ -115,6 +115,26 @@ TEST(SetHostAuto, NullValueApplied) {
     uriFreeUriMembersA(&uri);
 }
 
+TEST(SetHostAuto, NullValueAppliedNoPath) {
+    UriUriA uri = parseWellFormedUri("https://example.com");
+
+    EXPECT_EQ(uriSetHostAutoA(&uri, NULL, NULL), URI_SUCCESS);
+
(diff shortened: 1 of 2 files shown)


test/SetHostAuto.cpp
```diff
@@ -115,6 +115,26 @@ TEST(SetHostAuto, NullValueApplied) {
...
+
+TEST(SetHostAuto, NullValueAppliedNoPath2) {
+    UriUriA uri = parseWellFormedUri("https://example.com?foo=bar");
+
+    EXPECT_EQ(uriSetHostAutoA(&uri, NULL, NULL), URI_SUCCESS);
+
+    assertUriEqual(&uri, "https:?foo=bar");
+
+    uriFreeUriMembersA(&uri);
+}
```
hartwork: @TimWolla is this test adding any value that `NullValueAppliedNoPath` right above does not already add? I'm consdering to ask you top drop that second test but maybe there is an idea that I am missing?

TimWolla: > is this test adding any value that `NullValueAppliedNoPath` right above does not already add?

I added it to verify that “anything after the path” will not change the behavior. I can drop either of the tests if you'd like (or keep them both if you agree with the intention).

PR #298, opened 2026-04-26
[CVE-2026-42371] Fix range-length truncation in `CompareRange`

worwin:
This pull request fixes range-length truncation in CompareRange and adds a dedicated regression test.

The previous implementation converted pointer-difference lengths to int during comparison. On 64-bit platforms, this could truncate large range lengths and allow mismatched ranges to be treated as equal length before reaching content comparison.

The fix compares the full range lengths before content comparison.

The regression test exercises this case directly. I validated that the test crashes on the vulnerable revision and passes with this fix.

This PR contains two commits:
1. the fix
2. the regression test

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -314,6 +314,7 @@ if(URIPARSER_BUILD_TESTS)
     find_package(GTest 1.8.0 REQUIRED)
 
     add_executable(testrunner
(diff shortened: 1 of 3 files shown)


hartwork (approved):
@worwin thank you!

I'm about to fill the form at https://cveform.mitre.org/ just now…

hartwork:
Update: We received CVE-2026-42371 for this.

PR #304, opened 2026-05-04
[CVE-2026-44927] Stop truncating `ptrdiff_t` to `int`

hartwork:
CC @iliaal

### DIFF
--- a/src/UriCommon.c
+++ b/src/UriCommon.c
@@ -67,6 +67,7 @@
 
 #  include <assert.h>
 #  include <stddef.h>
+#  include <stdint.h>  // SIZE_MAX
 
 /*extern*/ const URI_CHAR * const URI_FUNC(SafeToPointTo) = _UT("X");
 /*extern*/ const URI_CHAR * const URI_FUNC(ConstPwd) = _UT(".");
@@ -129,8 +130,11 @@ bool URI_FUNC(RangeEquals)(const URI_TYPE(TextRange) * a, const URI_TYPE(TextRan
 UriBool URI_FUNC(CopyRange)(URI_TYPE(TextRange) * destRange,
                             const URI_TYPE(TextRange) * sourceRange,
                             UriMemoryManager * memory) {
-    const int lenInChars = (int)(sourceRange->afterLast - sourceRange->first);
-    const int lenInBytes = lenInChars * sizeof(URI_CHAR);
+    const size_t lenInChars = sourceRange->afterLast - sourceRange->first;
+    if (lenInChars > SIZE_MAX / sizeof(URI_CHAR)) {  // detect integer overflow
+        return URI_FALSE;
+    }
+    const size_t lenInBytes = lenInChars * sizeof(URI_CHAR);
     URI_CHAR * dup = memory->malloc(memory, lenInBytes);
     if (dup == NULL) {
         return URI_FALSE;
--- a/src/UriCommon.c
+++ b/src/UriCommon.c
@@ -173,7 +173,7 @@ UriBool URI_FUNC(RemoveDotSegmentsEx)(URI_TYPE(Uri) * uri, UriBool relative,
     walker->reserved = NULL; /* Prev pointer */
     do {
         UriBool removeSegment = URI_FALSE;
-        int len = (int)(walker->text.afterLast - walker->text.first);
+        const size_t len = walker->text.afterLast - walker->text.first;
         switch (len) {
         case 1:
             if ((walker->text.first)[0] == _UT('.')) {
--- a/src/UriQuery.c
+++ b/src/UriQuery.c
@@ -67,6 +67,7 @@
 
 #  include <limits.h>
 #  include <stddef.h> /* size_t */
+#  include <stdint.h>  // SIZE_MAX
 
 static int URI_FUNC(ComposeQueryEngine)(URI_CHAR * dest,
                                         const URI_TYPE(QueryList) * queryList,
@@ -267,8 +268,8 @@ UriBool URI_FUNC(AppendQueryItem)(URI_TYPE(QueryList) * *prevNext, int * itemCou
                                   const URI_CHAR * valueAfter, UriBool plusToSpace,
                                   UriBreakConversion breakConversion,
                                   UriMemoryManager * memory) {
-    const int keyLen = (int)(keyAfter - keyFirst);
-    const int valueLen = (int)(valueAfter - valueFirst);
+    const size_t keyLen = keyAfter - keyFirst;
+    const size_t valueLen = valueAfter - valueFirst;
     URI_CHAR * key;
     URI_CHAR * value;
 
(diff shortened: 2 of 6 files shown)


src/UriQuery.c
```diff
@@ -305,6 +318,11 @@ UriBool URI_FUNC(AppendQueryItem)(URI_TYPE(QueryList) * *prevNext, int * itemCou
 
     /* Fill value */
     if (valueFirst != NULL) {
+        // Detect integer overflow
+        if ((valueLen > SIZE_MAX - 1) || (valueLen + 1 > SIZE_MAX / sizeof(URI_CHAR))) {
+            return URI_FALSE;  // Raises malloc error
```
iliaal, line 328: I think `memory->free(memory, *prevNext);` and ` memory->free(memory, key);`  are needed here not to leak memory

src/UriQuery.c
```diff
@@ -285,6 +293,11 @@ UriBool URI_FUNC(AppendQueryItem)(URI_TYPE(QueryList) * *prevNext, int * itemCou
     }
     (*prevNext)->next = NULL;
 
+    // Detect integer overflow
+    if ((keyLen > SIZE_MAX - 1) || (keyLen + 1 > SIZE_MAX / sizeof(URI_CHAR))) {
+        return URI_FALSE;  // Raises malloc error
```
iliaal, line 300: I think memory->free(memory, *prevNext); is needed here not to leak memory

hartwork: @iliaal you're right — awesome! 🙏

src/UriQuery.c
hartwork: @iliaal you're right — awesome! 🙏

hartwork: @iliaal fix pushed

src/UriQuery.c
hartwork: @iliaal fix pushed

iliaal (approved):
LGTM

hartwork:
> LGTM

@iliaal many thanks for the review! 🙏

src/UriCommon.c
```diff
@@ -129,8 +130,11 @@ bool URI_FUNC(RangeEquals)(const URI_TYPE(TextRange) * a, const URI_TYPE(TextRan
 UriBool URI_FUNC(CopyRange)(URI_TYPE(TextRange) * destRange,
                             const URI_TYPE(TextRange) * sourceRange,
                             UriMemoryManager * memory) {
-    const int lenInChars = (int)(sourceRange->afterLast - sourceRange->first);
-    const int lenInBytes = lenInChars * sizeof(URI_CHAR);
+    const size_t lenInChars = sourceRange->afterLast - sourceRange->first;
```
TimWolla, line 133: Note: Inconsistent use of data types compared with `RangeEquals` (which uses ptrdiff_t). I'd say that `size_t` is more correct and would suggest adjusting `RangeEquals` for consistency.

src/UriRecompose.c
TimWolla, line 84: I think this static helper might be simpler to understand / verify if it used `size_t` internally with only the public `ToString()` and `ToStringCharsRequired()` doing the int <-> size_t conversion and overflow checking.

TimWolla (commented):
Consider adding a “checked multiplication” helper function to avoid the repeated manual overflow checks:

```
size_t alloc_size;
if (ckd_mul(&alloc_size, len, sizeof(URI_CHAR))) {
  return URI_FALSE;
}
memory->malloc(memory, alloc_size);
```

(I'm using C23’s `ckd_mul()` here for the example)

Edit: Fixed an incorrect negation of the return value.

hartwork:
> Consider adding a “checked multiplication” helper function to avoid the repeated manual overflow checks:
> 
> ```
> size_t alloc_size;
> if (!ckd_mul(&alloc_size, len, sizeof(URI_CHAR))) {
>   return URI_FALSE;
> }
> memory->malloc(memory, alloc_size);
> ```
> 
> (I'm using C23’s `ckd_mul()` here for the example)

@TimWolla interesting idea — thanks! I like that it removes some of the error prone details. The negation seems to be mistaken though: I find `true` documented as the case with an issue, not without. Will double check.

src/UriCommon.c
hartwork: @TimWolla why not — pull request #307 addresses this idea of yours

src/UriRecompose.c
hartwork: @TimWolla I consider this tied to the `int maxChars, int * charsWritten, int * charsRequired` in the function signature which serves public API that also uses ìnt`. If users had >2GB URIs, I would duplicate the API to a version with `size_t`, migrate the internal version to be all `size_t` and convert back and forth once. I'm unsure whether it's worth the effort. I'll sleep over it.

TimWolla:
> The negation seems to be mistaken though: I find `true` documented as the case with an issue, not without.

You are correct. I have fixed the above snippet. Luckily this would have been easy to notice: Virtually every multiplication would've failed, which is something the test suite should detect. 

How to best integrate something like that into the uriparser code base is something I leave up to you (e.g. if you only want to use a custom implementation, or if you want to conditionally rely on the C23 feature).

src/UriRecompose.c
TimWolla: The troubles I'm having with `int` also infecting the *internal* API is that the overflow checks need additional casts, which makes it much harder to verify them for correctness. That's why I suggested pushing the checks to the boundary. It should require fairly little code for the “public” API. Something like this:

```
int URI_FUNC(ToStringCharsRequired)(const URI_TYPE(Uri) * uri, int * charsRequired) {
    size_t charsRequiredInternal;
    int result = URI_FUNC(ToStringEngine)(NULL, uri, INT_MAX, NULL, &charsRequiredInternal);

    if (result != URI_SUCCESS) {
        return result;
    }

    // We're passing INT_MAX as `maxChars`, thus the required chars must fit an (int).
    assert(charsRequiredInternal < INT_MAX);

    *charsRequired = (int)charsRequiredInternal;

    return URI_SUCCESS;
}
```
```
int URI_FUNC(ToString)(URI_CHAR * dest, const URI_TYPE(Uri) * uri, int maxChars,
                       int * charsWritten) {
    size_t charsWrittenInternal;
    int result = URI_FUNC(ToStringEngine)(dest, uri, maxChars, &charsWrittenInternal, NULL);
    assert(charsWrittenInternal < INT_MAX);
    assert(charsWrittenInternal < maxChars);
    *charsWritten = (int)charsWrittenInternal;
    return result;
}
```

hartwork:
> > The negation seems to be mistaken though: I find `true` documented as the case with an issue, not without.
> 
> You are correct. I have fixed the above snippet. Luckily this would have been easy to notice: Virtually every multiplication would've failed, which is something the test suite should detect.

@TimWolla with test coverage, yes.


> How to best integrate something like that into the uriparser code base is something I leave up to you (e.g. if you only want to use a custom implementation, or if you want to conditionally rely on the C23 feature).

@TimWolla I have no plans of going beyond C99 with this code base. I will leave this as-is and sleep over it and the various ways to do or not do this in C99. The current code has one upside: The exposure of details makes it very clear what is happening.

src/UriRecompose.c
hartwork: @TimWolla the casts like `(size_t)INT_MAX` I am not concerned about personally, and I see no other casts in the diff for `ToStringEngine`. Adding `assert`s like these would allow crashing the process, so I would do error returns instead to not create a denial-of-service vulnerability. I may do that as a follow-up pull request with my workload is lower. I will sleep over it.

PR #306, opened 2026-05-04
Fix OOM related memory leak in `CopyUriMm`

hartwork:
CC @iliaal


iliaal (approved):
LGTM

P.S. Better then my naive version that missed the `destUri->pathHead != destWalker` check

hartwork:
@iliaal thanks for the review! :+1:

PR #310, opened 2026-05-13
Fix undefined behavior in `FilenameToUriString`

hartwork:
.. by no longer storing pointer `input - 1`.

CC @Smattr


Smattr:
LGTM!

hartwork:
@Smattr thank you! :+1:

PR #311, opened 2026-05-20
Protect `ToStringEngine` from integer overflow

hartwork:
CC @worwin


worwin (approved):
Retested against my original F-008 PoC. The vulnerable baseline returned success with a negative required-size value and triggered UBSan on the delimiter addition. With this patch, the same case now cleanly returns URI_ERROR_TOSTRING_TOO_LONG and no overflow is reported.

This fixes the reported issue. Approved.

hartwork:
@worwin thank you! :pray:

PR #312, opened 2026-05-20
Protect `AppendQueryItem` from item count integer overflow

hartwork:
CC @worwin


worwin (approved):
Retested with my original PoC for this issue. On the vulnerable code, the overflow reproduced. With this patch, the same case now fails safely instead of overflowing.

hartwork:
@worwin perfect, thank you! :pray:

PR #314, opened 2026-06-01
src/UriParse.c: Reset ranges in FreeUriMembersMm regardless of owner

alhudz:
Found by fuzzing the uriSet* API: a failed uriParseSingleUri leaves userInfo as a half-set range (first into the input, afterLast NULL) because FreeUriMembersMm only resets the range pointers when .owner is true. A later uriSetPath flips .owner via MakeOwner (which skips the malformed range) and the next free then frees that interior input pointer. Reset the range pointers unconditionally; only the free() stays gated by .owner.

### DIFF
--- a/src/UriParse.c
+++ b/src/UriParse.c
@@ -1951,47 +1951,53 @@ int URI_FUNC(FreeUriMembersMm)(URI_TYPE(Uri) * uri, UriMemoryManager * memory) {
 
     URI_CHECK_MEMORY_MANAGER(memory); /* may return */
 
-    if (uri->owner) {
-        /* Scheme */
-        if (uri->scheme.first != NULL) {
-            if (uri->scheme.first != uri->scheme.afterLast) {
-                memory->free(memory, (URI_CHAR *)uri->scheme.first);
-            }
-            uri->scheme.first = NULL;
-            uri->scheme.afterLast = NULL;
-        }
+    /* NOTE: The range pointers are reset to NULL regardless of .owner; only
+     *       the memory::free calls are gated by .owner.  A parse that stops
+     *       with an error funnels through here while .owner is URI_FALSE and
+     *       may have set .first of a range without .afterLast (e.g. a
+     *       tentative userInfo, see ParseOwnHostUserInfoNz), so leaving the
+     *       pointers untouched would keep a half-set range that a later
+     *       reuse (e.g. a uriSet* call followed by another free) frees as if
+     *       it were owned, freeing a pointer into the input. */
 
-        /* User info */
-        if (uri->userInfo.first != NULL) {
-            if (uri->userInfo.first != uri->userInfo.afterLast) {
-                memory->free(memory, (URI_CHAR *)uri->userInfo.first);
-            }
-            uri->userInfo.first = NULL;
-            uri->userInfo.afterLast = NULL;
-        }
+    /* Scheme */
+    if (uri->owner && (uri->scheme.first != NULL)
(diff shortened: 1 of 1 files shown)


hartwork:
@alhudz before I take a closer look at this, can you confirm that if there is any AI involved in this that you validated the whole thing as a human and that the diff has to be this big?

alhudz:
I reviewed and validated the changes manually. The diff is this size because of the scope of the changes, not any shortcut

hartwork:
@alhudz I had a first closer look now and from what I can see `FreeUriMembers`' job is not to make the URI structure reusable but only to free all memory associated to it, and it tries to do as little work as needed to do for that. So a call change of trying to parse, freeing and then calling `SetPath` does not seem legit and supported use of the API. If that's accurate, this pull requests makes things slower without fixing a bug. Can you help me understand what I am missing here?

alhudz:
The free in the middle isn't a caller free, it's the parser's own. `uriParseSingleUri` runs `FreeUriMembersMm` itself when a parse fails (`StopSyntax`, and again at the tail of `ParseSingleUriExMm`). So the caller sequence is just:

```c
UriUriA u; memset(&u, 0, sizeof u);
const char * s = "//$*c:.";
uriParseSingleUriA(&u, s, NULL);   /* fails, parser frees internally */
uriSetPathA(&u, s, s + 7);         /* MakeOwner flips .owner true */
uriFreeUriMembersA(&u);            /* invalid free, aborts */
```

No manual free in between. That last call aborts with `pointer being freed was not allocated`.

What goes wrong: on the userInfo path the parser sets `userInfo.first` into the input as USERINFO BEGIN (line 199) but errors out before USERINFO END, so `afterLast` never gets set. The failure cleanup then runs `FreeUriMembersMm` while `.owner` is still false, and the old code only reset the range pointers inside `if (uri->owner)`, so `userInfo` is left as `{ptr-into-input, NULL}`. The later `uriSetPath` calls `MakeOwner`, which skips that half-set range but flips `.owner` true, and the next free frees `userInfo.first`, a pointer into the caller's buffer.

So it's not reuse after a manual free. The parser's own failure cleanup is meant to leave the struct in the freed/clean state and it doesn't for that tentative range, which is what a normal build-from-scratch (`uriSet*` then free) then trips over.

On size: `userInfo` is the only range the failure path can leave half-set, so I'm happy to cut this down to just resetting `userInfo` unconditionally and leave the rest as-is if you'd prefer the minimal diff. Found it fuzzing the `uriSet*` API.

hartwork:
@alhudz thanks for elaborating!

> [T]he caller sequence is just:
> 
> ```c
> UriUriA u; memset(&u, 0, sizeof u);
> const char * s = "//$*c:.";
> uriParseSingleUriA(&u, s, NULL);   /* fails, parser frees internally */
> uriSetPathA(&u, s, s + 7);         /* MakeOwner flips .owner true */
> uriFreeUriMembersA(&u);            /* invalid free, aborts */
> ```

That's not legit use of the API: after the call to `uriParseSingleUriA` the caller has to check its return value and can only continue with that URI structure after a successful parse or would need to run a second parse to get the structure into a usable state. I believe we're at a point of "close as invalid" now. Can you confirm?

hartwork:
PS: `uriParseSingleUriA` does `memset(&u, 0, sizeof u);` already internally.

alhudz:
yeah, fair enough. it does hinge on calling `uriSet*` after a failed parse without re-checking the return value or re-parsing, which isn't supported use. no real bug on the supported path then. closing as invalid, thanks for taking the time to look.

PR #324, opened 2026-06-16
CMake: Allow CXX_STANDARD version to be settable

AdamMajer:
When trying to compile with newer GTest, it requires C++17 to be minimum. But since the C++ standard version is hardcoded, it's not possible to override it via -D.

If we use set with CACHE instead, the -DCMAKE_CXX_STANDARD=17 allows the version to be set when running cmake that is not overriden.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,9 +93,9 @@ if(URIPARSER_BUILD_TESTS OR URIPARSER_BUILD_FUZZERS)
     enable_language(CXX)
 
     if(URIPARSER_BUILD_FUZZERS)
-        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD 17 CACHE STRING "C++ Standard to use")
     else()
-        set(CMAKE_CXX_STANDARD 11)
+        set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ Standard to use")
     endif()
     set(CMAKE_CXX_STANDARD_REQUIRED ON)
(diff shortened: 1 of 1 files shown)


hartwork:
> When trying to compile with newer GTest, it requires C++17 to be minimum.

Hi @AdamMajer thanks for bringing this to attention! it seems to be official since GoogleTest 1.17.0 (of April 2025) according to https://github.com/google/googletest#release-1170 .

That probably means that the CI should start covering that and an older version of GoogleTest.

> But since the C++ standard version is hardcoded, it's not possible to override it via -D.

I see.

> If we use set with CACHE instead, the -DCMAKE_CXX_STANDARD=17 allows the version to be set when running cmake that is not overriden.

I'll need to sleep over that approach. Are there alternatives that you already thought of and evaluated?

AdamMajer:
There's not really an alternative here. If I set CXXFLAGS, this is prepended, and the -std=c++11 is appended and so overrides what I set via env.

The change here seems like the simplest change, allowing the standard version to be set in the CMakeLists.txt as default and expected, but allow users to override if necessary.

hartwork:
@AdamMajer I am going forward now with an unconditional bump to C++17 in pull request #325. I'm assuming this alternative works well for your scenario also. Please report back how well it does. Thanks! :pray:

hartwork:
Closing as superseded by #325 which addresses the original problem but in a way where (1) things now work out of the box and (2) no new interface is being introduced. @AdamMajer thanks for reporting this issue upstream!

PR #325, opened 2026-06-17
Fix support for GoogleTest >=1.17.0 / Make CI cover (both less and) more recent GoogleTest (alternative to #324)

hartwork:
Alternative to #324

Related:
- https://github.com/google/googletest/issues/4731
- https://github.com/google/googletest/pull/4877

CC @AdamMajer @charles-lunarg

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -93,9 +93,9 @@ if(URIPARSER_BUILD_TESTS OR URIPARSER_BUILD_FUZZERS)
     enable_language(CXX)
 
     if(URIPARSER_BUILD_FUZZERS)
-        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD 17)  # for use of Clang's fuzzer/FuzzedDataProvider.h
     else()
-        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD 17)  # for GoogleTest >=1.17.0
     endif()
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -93,9 +93,9 @@ if(URIPARSER_BUILD_TESTS OR URIPARSER_BUILD_FUZZERS)
     enable_language(CXX)
 
     if(URIPARSER_BUILD_FUZZERS)
-        set(CMAKE_CXX_STANDARD 17)
+        set(CMAKE_CXX_STANDARD 17)  # for use of Clang's fuzzer/FuzzedDataProvider.h
     else()
-        set(CMAKE_CXX_STANDARD 11)
+        set(CMAKE_CXX_STANDARD 17)  # for GoogleTest >=1.17.0
```
charles-lunarg, line 98: Pardon my ignorance, but this seems like a pointless if-else branch now. It'll always run `set(CMAKE_CXX_STANDARD 17)` unless I'm missing something.

hartwork: @charles-lunarg that was actually a conscious decision. I am optimizing for a clearer diff and the expectation that the two will diverge again in the future here. Thanks for asking! :+1:

charles-lunarg: That makes sense on a line-by-line diff, just looked odd with the per-commit diff the PR has. Thanks for the explanation.

Issue #315, opened 2026-06-02
uriparser 1.0.2 build error with glibc <2.29


satmandu:
On our x86_64 Linux/chromebrew systems we are seeing this build error:

```
UriMemory.c:109:12: error: returning ‘int’ from a function with return type ‘void *’ makes pointer from integer without a cast [-Wint-conversion]
```

hartwork:
Hi @satmandu I believe that means that the function `reallocarray` is considered present by the build system but absent by the compiler. Here's the pre-processor parts to it:

https://github.com/uriparser/uriparser/blob/a5763fa2d6d8e606e3ab5313871c2837fadad23f/src/UriMemory.c#L48-L66

May I ask what libc and version of that you have in ChromeOS and/or chromebrew? Can you help me debug the cause for this?

satmandu:
> Hi @satmandu I believe that means that the function `reallocarray` is considered present by the build system but absent by the compiler. Here's the pre-processor parts to it:
> 
> https://github.com/uriparser/uriparser/blob/a5763fa2d6d8e606e3ab5313871c2837fadad23f/src/UriMemory.c#L48-L66
> 
> May I ask what libc and version of that you have in ChromeOS and/or chromebrew? Can you help me debug the cause for this?

We tend to use glibc 2.23 for `i686`, and glibc 2.27 for both `x86_64` and `armv7l`.

We use the oldest glibc possible to have maximum compatibility with EOL devices.

satmandu:
Also, we are currently using GCC 16.1 as our default compiler.

hartwork:
@satmandu thanks for elaborating!

glibc 2.27 is from 2018 and 2.23 from 2016 which is both way older than what I intend to support, and with my security hat on that use of glibc sounds a bit worrisome to me on paper, given that it would take to backport all security fixes [including those that were never labeled as security upstream]((https://blog.hartwork.org/posts/how-much-security-is-in-long-term-support/)) 🤔 

I think my options are:
- a) fix the pre-processor guards — likely by adding `_GNU_SOURCE` — (or the overeager build system) and violate my 5 years support window
- b) ask you to fix the build on your side because unsupported version of glibc
- c) add explicit guards for glibc from <=5 years ago plus (c) to make the issue more obvious

I'm not too happy with any of these. Let me sleep over this. I am happy about feedback in the meantime.

hartwork:
@satmandu I thought about this more. I see multiple versions of glibc packaged in chromebrew…

[...]

PR #327, opened 2026-07-02
reallocarray: make use optional

akallabeth:
If detection fails allow overriding it via CMake option

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,7 +141,10 @@ endif()
 # UriConfig.h
 #
 check_symbol_exists(wprintf wchar.h HAVE_WPRINTF)
-check_function_exists(reallocarray HAVE_REALLOCARRAY)  # no luck with CheckSymbolExists
+option(WITH_REALLOCARRAY "Try to detect reallocarray" ON)
+if (WITH_REALLOCARRAY)
+    check_function_exists(reallocarray HAVE_REALLOCARRAY)  # no luck with CheckSymbolExists
+endif()
 configure_file(${CMAKE_CURRENT_SOURCE_DIR}/src/UriConfig.h.in UriConfig.h)
 
 #


hartwork (changes requested):
Hi @akallabeth, what is the problem that this pull request solves and for which platform (compiler, operating system, hardware)?

hartwork:
@akallabeth PS: Did you see #320 and does it solve your problem?

akallabeth:
@hartwork cross compile to `iOS` (as dependency of https://github.com/freerdp/freerdp )
the detection falsely detects support, writes it to the header and fails at compile/link time

hartwork:
@akallabeth sounds like issue #315 that PR #320 was fixing: `master` has that fixed already. Can you confirm?

akallabeth:
@hartwork build from 6d7db1bac2fa4a712c35789b542388510aefabe9 not fixed

hartwork:
@akallabeth more details please for a full bug report. Thanks!

hartwork:
@akallabeth PS: e.g. which target libc, which version, which compiler, which version, precise compiler output etc

hartwork:
> @hartwork cross compile to `iOS` (as dependency of https://github.com/freerdp/freerdp ) the detection falsely detects support, writes it to the header and fails at compile/link time

@akallabeth PS: This ^^ means that CMake and the compiler do no agree on whether `reallocarray` is available, with CMake saying "yes" and the compiler saying "no". Uncovered why in detail would be most helpful and be an actual fix. Assuming that your target platform is recent enough to be supported. My target support window is "relevant platforms and 5 years".

akallabeth:
@hartwork well, current `Xcode` and most up to date `iOS SDK`

hartwork:
@akallabeth that's good. Is my understanding correct that it does not have `reallocarray` at all? The question then would be why CMake says it could be found. Could you check the logs in the `CMakeFiles` folder for anything related?

hartwork:
@akallabeth I was hoping this would show up with macOS in CI also but it's fine, see #329. Any insights from your `CMakeFiles/CMakeConfigureLog.yaml` regarding section `"Looking for reallocarray"`?

akallabeth:
* [CMakeConfigureLog.yaml.tar.gz](https://github.com/user-attachments/files/29622434/CMakeConfigureLog.yaml.tar.gz)
* tried to build with `cmake -GXcode -Bxxx -S. -DCMAKE_TOOLCHAIN_FILE=../freerdp/cmake/ios.toolchain.cmake -DPLATFORM=SIMULATORARM64 -DDEPLOYMENT_TARGET=15.0`
* toolchain file from https://github.com/leetal/ios-cmake.git

hartwork:
@akallabeth thanks for sharing! I don't see anything revealing in the YAML output. When I play with this `CMakeLists.txt`…

```CMake
cmake_minimum_required(VERSION 3.10)

project(debug123)

include(CheckFunctionExists)

check_function_exists(reallocarray HAVE_REALLOCARRAY)

message(STATUS "HAVE_REALLOCARRAY is [${HAVE_REALLOCARRAY}]")
```

… by running …

```bash
cmake --debug-trycompile .
find /tmp/ -name CheckFunctionExists.c -print -exec cat {} \; 2>/dev/null
```

… locally on my Linux box, I get to see this C code for the check:

```c
#ifdef CHECK_FUNCTION_EXISTS

#  ifdef __cplusplus
extern "C"
#  endif
  char
  CHECK_FUNCTION_EXISTS(void);
#  ifdef __CLASSIC_C__
int main()
{
  int ac;
  char* av[];
#  else
int main(int ac, char* av[])
{
#  endif
  CHECK_FUNCTION_EXISTS();
  if (ac > 1000) {
    return *av[0];
  }
  return 0;
}

#else /* CHECK_FUNCTION_EXISTS */

#  error "CHECK_FUNCTION_EXISTS has to specify the function"
```

(It is compiled with `-DCHECK_FUNCTION_EXISTS=reallocarray` by CMake.)

My guess is that the `char CHECK_FUNCTION_EXISTS(void);` is maybe causing the false positive but it would not explain how the link stage gets to succeed.

I guess we are dealing with some issue with either CMake itself or that particular toolchain file here. (CMake itself is a possibility given e.g. https://github.com/libexpat/libexpat/blob/0ac39627c7942f760e27cad67c8a6f09864381fe/expat/ConfigureChecks.cmake#L108-L113).

What we can try is me converting the check to use [check_c_source_compiles](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) instead and you checking if that works any better in your environment. If things go well, I'll have a draft pull request in a few minutes.

hartwork:
@akallabeth PS: The `-DENABLE_STRICT_TRY_COMPILE=(BOOL) - Disabled by default` at https://github.com/leetal/ios-cmake#additional-options caught my eye. Is that something you tried? Could it have an effect on your case?

akallabeth:
@hartwork ok, you found it.
compiling with `-DENABLE_STRICT_TRY_COMPILE=ON` works.

hartwork:
> What we can try is me converting the check to use [check_c_source_compiles](https://cmake.org/cmake/help/latest/module/CheckCSourceCompiles.html) instead and you checking if that works any better in your environment. If things go well, I'll have a draft pull request in a few minutes.

@akallabeth ready to try now if you find a minute: #331

hartwork:
> @hartwork ok, you found it. compiling with `-DENABLE_STRICT_TRY_COMPILE=ON` works.

@akallabeth interesting! Could you try #331 without it (and CMake cache deleted) to see if the PR improves on the original situation?

akallabeth:
> > @hartwork ok, you found it. compiling with `-DENABLE_STRICT_TRY_COMPILE=ON` works.
> 
> @akallabeth interesting! Could you try #331 without it (and CMake cache deleted) to see if the PR improves on the original situation?

done and works as well.

hartwork:
@akallabeth thanks for trying it out! :pray:

PR #331, opened 2026-07-03
CMake: Migrate check for `reallocarray` to `check_c_source_compiles`

hartwork:
.. to avoid false positives, e.g. with XCode and iOS.

In reaction to https://github.com/uriparser/uriparser/pull/327#issuecomment-4866038009

CC @akallabeth


akallabeth (approved):
Nice work, this fixes the detection issue for iOS here.

Issue #335, opened 2026-07-06
Normalize IPv6 addresses when normalizing URIs


TimWolla:
I patched `tool/uriparse.c` as follows:

```
--- i/tool/uriparse.c
+++ w/tool/uriparse.c
@@ -104,6 +104,8 @@ int main(int argc, char * argv[]) {
                     state.errorPos, (long unsigned int)(state.errorPos - argv[i]));
             retval = EXIT_FAILURE;
         } else {
+            uriNormalizeSyntaxA(&uri);
+
             if (uri.scheme.first) {
                 dumpRange("scheme", &(uri.scheme));
             }
@@ -146,6 +148,23 @@ int main(int argc, char * argv[]) {
                 printf("%*s%s\n", (int)strlen(absolutePathLabel), "",
                         "(always false for URIs with host)");
             }
+
+            char * uriString;
+            int charsRequired;
+            if (uriToStringCharsRequiredA(&uri, &charsRequired) != URI_SUCCESS) {
+                /* Failure */
+            }
+            charsRequired++;
+            uriString = malloc(charsRequired * sizeof(char));
+            if (uriString == NULL) {
+                /* Failure */
+            }
+            if (uriToStringA(uriString, &uri, charsRequired, NULL) != URI_SUCCESS) {
+                /* Failure */
+            }
+
+            printf("Recomposed: %s\n", uriString);
+            free(uriString);
         }
         printf("\n");
```

When I then run: `./uriparse 'http://[2001:db8::0:0:0123]/'` I get the following output:

```
uri:          http://[2001:db8::0:0:0123]/
scheme:       http
hostText:     2001:db8::0:0:0123
hostData.ip6: 2001:db8::123
 .. pathSeg:  
absolutePath: false
              (always false for URIs with host)
Recomposed: http://[2001:0db8:0000:0000:0000:0000:0000:0123]/
```

I would expect two things:

1. The recomposed URL should be in sync with `hostText`.
2. When I run `uriNormalizeSyntaxA()`, I expect the `hostText` to be “minimal in length” afterwards (so matching the `hostData.ip6` output that was generated by `inet_ntop`.

see php/php-src#22618. The `getHost()` getter reads `.hostText`, whereas `toString()` uses `uriToStringA()` internally.

hartwork:
Hi @TimWolla,

[...]

PR #336, opened 2026-07-06
src/UriNormalize.c: Normalize IPv6 addresses

TimWolla:
This is a proof-of-concept for #335.


hartwork:
@TimWolla thanks for the PoC. I will not look into this in more detail before August.

prateekbhujel:
I tested this against the php-src reproducer and found one edge case in the zero run selection. `2001:db8:0:1:1:1:1:1` currently becomes `2001:db8::1:1:1:1:1`, but RFC 5952 section 4.2.2 says `::` must not shorten a single 16-bit zero field.

I opened TimWolla/uriparser#1 with the fix and tests. It also makes the normalization helper compare the recomposed text, since `uriEqualsUriW()` only compares the 16-byte IPv6 value. All 412 tests pass locally.

PR #337, opened 2026-07-13
Add support for percent encoding of query strings according to RFC 3986

kocsismate:
The `URI_FUNC(Escape)` and `URI_FUNC(EscapeEx)` functions are intended to percent-encode the input according to the application/x-www-form-urlencoded media type as defined by RFC 1866:

> 1. The form field names and values are escaped: space characters are replaced by `+', and then reserved characters are escaped as per [URL]; that is, non-alphanumeric characters are replaced by `%HH', a percent sign and two hexadecimal digits representing the ASCII code of the character. Line breaks, as in multi-line text field values, are represented as CR LF pairs, i.e. `%0D%0A'."

However, RFC 3986 has a different syntax:
> query = *( pchar / "/" / "?" )

First of all, it allows the following reserved characters to be present without percent-encoding: "/", "?", "@", and ":". Then it percent-encodes the " " as "%20", and finally, it doesn't perform line-break normalization.

That's why the newly added `URI_FUNC(EscapeQueryRfc3986)` and `URI_FUNC(EscapeQueryRfc3986Ex)` functions add support for strict RFC 3986 percent-encoding of the query string.

And last but not least, the `URI_FUNC(EscapeQueryRfc1866)` and `URI_FUNC(EscapeQueryRfc1866Ex)` functions are added for convenience to implement strict RFC 1866 percent-encoding of the form data - whereas the original `URI_FUNC(Escape)` and `URI_FUNC(EscapeEx)` functions have the `spaceToPlus` and `normalizeBreaks` parameters which allows a hybrid behavior.


hartwork:
@kocsismate thanks for the pull request. I will not look into this in more detail before August.

kocsismate:
That's fine, enjoy the summer! 🙂

src/UriEscape.c
```diff
@@ -154,6 +150,26 @@ URI_CHAR * URI_FUNC(EscapeEx)(const URI_CHAR * inFirst, const URI_CHAR * inAfter
             prevWasCr = URI_TRUE;
             break;
 
+        case URI_SET_SUB_DELIMS(_UT):
+            /* intentional fallback */
```
kocsismate, line 154: in the PHP project, we have a dedicated macro to suppress the `this statement may fall through [-Werror=implicit-fallthrough=]` errors:

```
#if __STDC_VERSION__ >= 202311L || defined(__cplusplus)
# define ZEND_FALLTHROUGH [[fallthrough]]
#elif defined(__GNUC__) && __GNUC__ >= 7
# define ZEND_FALLTHROUGH __attribute__((__fallthrough__))
#else
# define ZEND_FALLTHROUGH ((void)0)
#endif
```

So we should either use something similar here  or I can also rewrite the code so that each of the relevant `case`s call the same function.

TimWolla: This kind of fallthrough indicator is typically only used when the `case` is non-empty. `case 'A': case 'B': something();` is fine.

